PR #1, opened 2019-06-07
Some updates after running against bmcs

ncode:
Hello,

Thanks a lot for your project. It's the first amiable implementation of redfish for go that I get to see! I've added some small bits, after running it against dell, hp and supermicro bmcs. I'm also planning on working in the authentication part if you don't mind, because it's required for integrating on our project https://github.com/bmc-toolbox/bmclib.

### DIFF
--- a/school/service.go
+++ b/school/service.go
@@ -41,7 +41,9 @@ type Service struct {
 // UnmarshalJSON unmarshals a Service object from the raw JSON.
 func (s *Service) UnmarshalJSON(b []byte) error {
 	type temp Service
-	type linkReference struct {
+	var t struct {
+		temp
+		// Link references
 		Chassis            common.Link
 		Managers           common.Link
 		TaskService        common.Link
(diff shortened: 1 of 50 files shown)


stmcginnis:
Awesome, thanks @ncode !

Quick first pass through looks good, I just want to do a little local testing first. Thanks for contributing!

school/service.go
```diff
@@ -67,32 +65,35 @@ func (s *Service) UnmarshalJSON(b []byte) error {
...
-	s.taskService = string(t.Links.TaskService)
-	s.sessionService = string(t.Links.SessionService)
-	s.storageServices = string(t.Links.StorageServices)
-	s.storageSystems = string(t.Links.StorageSystems)
-	s.accountService = string(t.Links.AccountService)
-	s.eventService = string(t.Links.EventService)
-	s.registries = string(t.Links.Registries)
-	s.systems = string(t.Links.Systems)
-	s.compositionService = string(t.Links.CompositionService)
+	s.chassis = string(t.Chassis)
```
stmcginnis: Good catch. It looks like the Redfish-Interface-Emulator code currently has a bug where it has these properties under Links, while according to the ServiceRoot spec they should be root properties. I've filed https://github.com/DMTF/Redfish-Interface-Emulator/issues/67 to fix this in the emulator code.

PR #2, opened 2019-07-09
Adding support for authentication via session

ncode:
Hello @stmcginnis 

I've started implementing the authentication via session to gofish. I've added some methods to redfish/session.go and also to services.go using it as a proxy (or wrapper), following the same pattern you've done for the List methods. I will probably finish the session creation today or tomorrow, so I would like to know if you have any preference on how I should organize the code. 

Thanks!!

[]'s

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,25 @@ package main
 import (
     "fmt"
 
-    "github.com/stmcginnis/gofish/school"
+    gofish "github.com/stmcginnis/gofish/school"
 )
 
 func main() {
-    c := gofish.APIClient("localhost", 5000, false)
-    service, _ := school.ServiceRoot(c)
+    c, err := gofish.APIClient("http://localhost:5000", nil)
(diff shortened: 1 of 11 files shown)


stmcginnis:
Fantastic! Thanks for working on this @ncode.

Everything looks good to me so far. Feel free to organize things how you think would be best and we can take a look through the final results before merging.

ncode:
Hi @stmcginnis,  

Now we able able to create session and use them to issue requests with the token. I've also added two examples on how to use it. Let me know what you think and so that I can start working on reboot, power cycle, pxe and other common bmc actions.

Thanks!

stmcginnis:
This looks great. I don't have a system I can test against now, but I trust you've been doing that anyway, so works for me. ;)

Some minor things - we do want license headers even in the example files. I can get that added in a follow up though. Thanks for adding those examples, that is really great!

PR #4, opened 2019-07-12
Remove Default.*Path and update ComputerSystem 

ncode:
I've also started testing the code against a few different vendors and multiple versions of RedFish to ensure compatibility and also that I won't break anything :).

The only tricky part for me is to test the swordfish side, because I don't have any device capable a this moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,14 +21,25 @@ package main
 import (
     "fmt"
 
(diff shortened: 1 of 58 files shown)


ncode:
this patch looks messy, met me fix it first

PR #5, opened 2019-07-12
Remove Default.*Path and update ComputerSystem to match the redfish spec

ncode:
I've also started testing the code against a few different vendors and multiple versions of RedFish to ensure compatibility and also that I won't break anything :).

These changes have been tested against:
      // HP Gen10
      // Supermicro X10
      // Dell M630
      // Intel S2600WF0
      // Dell M640

The only tricky part for me is to test the swordfish side, because I don't have any device capable a this moment.

----

Adds Actions and create a list of possible ResetType

It's now possible to use the target from Actions to issue a reboot
action. Next step is to work on the boot once action

Update computersystem types to unpack as expected

Dell M640s also require the Accept Header, not only the Content-Type

### DIFF
--- a/school/redfish/computersystem.go
+++ b/school/redfish/computersystem.go
@@ -18,10 +18,6 @@ import (
 	"github.com/stmcginnis/gofish/school/common"
 )
 
-// DefaultComputerSystemPath is the default URI for the ComputerSystem
-// object.
-const DefaultComputerSystemPath = "/redfish/v1/ComputerSystem"
-
 // BootOrderTypes is the choice of boot order property to use when controller
 // the persistent boot order for this computer system.
 type BootOrderTypes string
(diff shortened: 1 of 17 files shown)


school/redfish/computersystem.go
```diff
@@ -300,6 +294,55 @@ type Boot struct {
...
+
+	return nil
+}
+
+// Reset describe the type off reset to be issue by the resource
+type ResetType string
+
+const (
+	// On shall be used to power on the machine
+	On ResetType = "On"
```
stmcginnis: Since there's some overlap in enum names, I've been following the convention of appending the type name to the name to make it clear and to prevent conflicts.

I think it would be good to keep following this convention, so that would make these OnResetType, ForceOffResetType, etc. A little more verbose, but at least in a couple cases it was necessary.

school/redfish/computersystem.go
```diff
@@ -565,7 +626,7 @@ type MemorySummary struct {
 	Status common.Status
 	// TotalSystemMemoryGiB is the amount of configured system general purpose
 	// volatile (RAM) memory as measured in gibibytes.
-	TotalSystemMemoryGiB int
+	TotalSystemMemoryGiB float32
```
stmcginnis: Ah, I was wondering if that was going to be an issue. ;)

There are probably more of these cases where we'll have to watch out for places that need updating.

Issue #11, opened 2019-07-24
Data type for ReadingCelsius


jenningsloy318:
Hi 
I found that the type of `ReadingCelsius` may be not suitable.
I get the return value from Lenovo SR850, the value is a `"ReadingCelsius": -42.8`, it should be a float

```json
....
{
      "UpperThresholdFatal": null,
      "PhysicalContext": "CPU",
      "LowerThresholdFatal": null,
      "UpperThresholdCritical": null,
      "RelatedItem": [
        {
          "@odata.id": "/redfish/v1/Systems/1/"
        },
        {
          "@odata.id": "/redfish/v1/Chassis/1/"
        },
        {
          "@odata.id": "/redfish/v1/Systems/1/Processors/4"
        }
      ],
      "Status": {
        "State": "Enabled"
      },
      "SensorNumber": 238,
      "Name": "CPU 4 DTS",
      "MaxReadingRangeTemp": 0,
      "UpperThresholdNonCritical": null,
      "@odata.id": "/redfish/v1/Chassis/1/Thermal/#/Temperatures/7",
      "@odata.type": "#Thermal.v1_3_0.Temperature",
      "MinReadingRangeTemp": -51,
      "MemberId": "CPU_4_DTS",
      "LowerThresholdNonCritical": null,
      "LowerThresholdCritical": null,
      "ReadingCelsius": -42.8
    },
....
```

 

```go
		chassisThermal, err := chasssis.Thermal()
		if err !=nil {
			log.Fatalf("Errors Getting  Thermal from chassis : %s",err)
		}
```
so I got following error 
```
Errors Getting   Thermal from chassis : json: cannot unmarshal number -42.8 into Go struct field .ReadingCelsius of type int  source="chassis_collector.go:164"
``` 
Created a pull request for this #12

PR #12, opened 2019-07-24
change data type of the chasis thermal and power

jenningsloy318:
Closes: #11

### DIFF
--- a/school/redfish/chassis.go
+++ b/school/redfish/chassis.go
@@ -181,15 +181,15 @@ type ThermalInfo struct {
 		Name                      string
 		SensorNumber              int
 		Status                    common.Status
-		ReadingCelsius            int
-		UpperThresholdNonCritical int
-		UpperThresholdCritical    int
-		UpperThresholdFatal       int
-		LowerThresholdNonCritical int
-		LowerThresholdCritical    int
-		LowerThresholdFatal       int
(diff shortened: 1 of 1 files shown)


school/redfish/computersystem.go
```diff
@@ -363,7 +363,7 @@ type ComputerSystem struct {
 	AssetTag string
 	// bios shall be a link to a resource of
 	// type Bios that lists the Bios settings for this system.
-	bios string
+	bios common.Link
```
stmcginnis: It would be preferable to keep these strings and follow the pattern used in other object definitions to have the extraction handled in the JSON decoding.

jenningsloy318:
HI @stmcginnis 
I know your point, but when I use your code, I got following error 

```
FATA[0013] Errors Getting systems from service : json: cannot unmarshal object into Go struct field .NetworkInterfaces of type string  "
FATA[0122] Errors Getting systems from service : json: cannot unmarshal object into Go struct field CSLinks.CooledBy of type string  "
````

so keep the type as string, there must other changes need to be adjusted

stmcginnis:
That error is different than the changes that were proposed. Looks like there are a few more properties that need to be made private. I should probably be able to get to that in the next couple days to better flesh out the ComputerSystem object.

Thanks for the details!

PR #14, opened 2019-07-25
modify poweredBy and CooledBy to common.link, as they are both links

jenningsloy318:
to fix the error

```
FATA[0122] Errors Getting systems from service : json: cannot unmarshal object into Go struct field CSLinks.CooledBy of type string 
FATA[0122] Errors Getting systems from service : json: cannot unmarshal object into Go struct field CSLinks.PoweredBy of type string 
```


stmcginnis:
Thanks!

PR #29, opened 2019-08-15
Fix various type errors with Thermal, Redundancy, and Power

### LINKED ISSUE #28: Type errors and missing items
Hi,

from real test, found found type errors
1. power votage type error, most of them are `float` not `int`, so need to change to `float32`
2. power metrcs also are float for watts, so also need to change to `float32`
3. Redundancy. RedundancySet should be `common.Links` not string 
4.  not sure why `common.Entity`  is removed from `chassis` objects, such as temperature/fan/power....

stmcginnis:
This corrects a few property definitions to make them match the schema.

Closes #28

### DIFF
--- a/school/redfish/power.go
+++ b/school/redfish/power.go
@@ -92,19 +92,19 @@ type InputRange struct {
 	InputType InputType
 	// MaximumFrequencyHz shall contain the value in Hertz of the maximum line
 	// input frequency which the power supply is capable of consuming for this range.
-	MaximumFrequencyHz int
+	MaximumFrequencyHz float32
 	// MaximumVoltage shall contain the value in Volts of the maximum line input
 	// voltage which the power supply is capable of consuming for this range.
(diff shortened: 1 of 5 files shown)


mike-pittelko:
Would it be helpful to use one of the more robust json->go tools?
https://pocketgophers.com/go-structs-from-json/   (multiple tools listed here)

stmcginnis:
I did think about that, but the issue is actually that the json files aren't rich enough to represent some things.

The actual source of truth for the redfish and swordfish schema are a set of XML files in private repos under DMTF and SNIA. That XML has a lot more clarity on object types and relationships. Unfortunately, they are private to only members of those orgs, so not the best source for an open source project to use.

I think most objects are now defined in the code, though not quite all. If it looks like we need to generate the definitions for a bunch more objects, then maybe we can look at updating the helper scripts.

PR #36, opened 2019-09-07
Adds NewAPIClient method

joelrebel:
The change merged in 33f8f279b8ddcac08b02472235aed76d56d3267e removes the `gofish.APIClient` method which would allow a user to pass in his own `net/http.Client`                                                                                                 
                                                                                                                                                                                                                                                                               
This change prevents the user from being able to tune the transport parameters - something we've seen as necessary in [bmclib](https://github.com/bmc-toolbox/bmclib/blob/master/internal/httpclient/httpclient.go#L15)                                                        
                                                                                                                                                                                                                                                                               
I would like to propose this change, that add the `NewAPIClient` method which gives the user of this library a bit more flexibility in the connection setup.                                                                                                               

Do suggest alternatives if this does not seem ideal :)

### DIFF
--- a/client.go
+++ b/client.go
@@ -85,10 +85,46 @@ func Connect(config ClientConfig) (c *APIClient, err error) {
 		}
 
 		client.Service = service
-		client.auth = auth
+		client.Auth = auth
 	}
 
 	return client, err
+}
+
+// NewAPIClient just returns a new APIClient, without a connection,
+// note: APIClient.Auth, APIClient.Service are left to be set by the caller
+func NewAPIClient(endpoint string, insecure bool, httpClient *http.Client) (c *APIClient, err error) {
+
+	if !strings.HasPrefix(endpoint, "http") {
+		return c, fmt.Errorf("endpoint must starts with http or https")
+	}
+
+	c = &APIClient{endpoint: endpoint}
+
+	// setup http client with default transport params
+	if httpClient == nil {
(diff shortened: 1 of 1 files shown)


stmcginnis:
Thanks @joelrebel. Sorry, I had hoped to follow up with a PR to bmclib to adjust for the breaking changes I made, but I ran out of time.

Can you explain a little more about what kinds of controls you need? It's likely others will need them too, so I'd rather build in a mechanism into the gofish client rather than requiring that the consumer of the library knows to do some other external setup steps first to get around client limitations.

client.go
```diff
@@ -85,12 +85,48 @@ func Connect(config ClientConfig) (c *APIClient, err error) {
...
-		client.auth = auth
+		client.Auth = auth
 	}
 
 	return client, err
 }
 
+// NewAPIClient just returns a new APIClient, without a connection,
+// note: APIClient.Auth, APIClient.Service are left to be set by the caller
+func NewAPIClient(endpoint string, insecure bool, httpClient *http.Client) (c *APIClient, err error) {
```
stmcginnis: I think if we do need to have a way to pass in an existing http.Client, it would be better to just add a field to the ClientConfig struct where a user can provide it there and use the standard Connect(config) call rather than a variation like this.

joelrebel: That sounds right, I'll close this PR and open another one based on your suggestion.

PR #37, opened 2019-09-07
client: allow an http.Client to be passed in

joelrebel:
This is based on https://github.com/stmcginnis/gofish/pull/36#discussion_r321977765


stmcginnis:
Thanks Joel!

PR #38, opened 2019-09-11
Adds Put and Patch interface, also SetBoot and Reset for ComputerSystem

ncode:
This PR makes possible to set Boot options, Boot Overrides and Reboot actions via Gofish.

### DIFF
--- a/redfish/computersystem.go
+++ b/redfish/computersystem.go
@@ -694,6 +694,32 @@ func (computersystem *ComputerSystem) SetBoot(b Boot) (bool, error) {
 	return true, nil
 }
 
+// Reset This action shall perform a reset of the ComputerSystem.  For systems which implement ACPI Power Button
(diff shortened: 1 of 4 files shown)


redfish/computersystem.go
```diff
@@ -635,6 +671,55 @@ func (computersystem *ComputerSystem) SecureBoot() (*SecureBoot, error) {
...
+		return false, err
+	}
+
+	resp, err := computersystem.Client.Post(computersystem.Actions.ComputerSystemReset.Target, payload)
+	if err != nil {
+		return false, err
+	}
+	defer resp.Body.Close()
+
+	return true, nil
```
stmcginnis: Is there a reason for wanting a bool returned as well rather than just the normal check of "if err != nil"?

stmcginnis:
Thanks @ncode !

PR #50, opened 2019-12-30
fixed typos

richardelling:
fixed typos, mostly in comments


stmcginnis:
Thanks! I was hesitant to do this before since rerunning the code generation on new version of the spec would just pull back in the same typos from the spec files, but I think things have changed enough at this point that I'll need to look at better ways of pulling in updates anyway.

PR #51, opened 2019-12-30
style and minor fixes

richardelling:
* fix golang and PEP8 style complaints
* cap() is a golang builtin
* error handling for dumpwriter writes


stmcginnis:
Thanks @richardelling, looks good.

Issue #49, opened 2019-12-30
Unable to list BIOS attributes?


goldyfruit:
Hi guys,

I'm trying to list BIOS attribute but it seems that I'm missing something.

```
	config := gofish.ClientConfig{
		Endpoint: "https://" + credentials.DriverInfo.IpmiAddress,
		Username: credentials.DriverInfo.IpmiUsername,
		Password: credentials.DriverInfo.IpmiPassword,
		Insecure: true,
	}
	c, err := gofish.Connect(config)
	if err != nil {
		fmt.Print(err)
	}
	service := c.Service
	systems, _ := service.Systems()

	for _, system := range systems {
		bios, _ := system.Bios()
		fmt.Print(bios)
	}
```
I got this output but no attributes.
```
&{{Bios BIOS Configuration Current Settings 0xc000dc6870} /redfish/v1/$metadata#Bios.Bios  /redfish/v1/Systems/System.Embedded.1/Bios #Bios.v1_0_2.Bios BiosAttributeRegistry.v1_0_0 BIOS Configuration Current Settings /redfish/v1/Systems/System.Embedded.1/Bios/Actions/Bios.ChangePassword /redfish/v1/Systems/System.Embedded.1/Bios/Actions/Bios.ResetBios}
```
How can I display the content of `BIOS Configuration Current Settings` ?
Using a simple `curl` display what I expect.

stmcginnis:
This is one area I need to address yet. The attributes are a loose set of typed values, so I just need to think about how best to handle this with go.

I'm actually hoping to spend some time on gofish today, so hopefully I will have an update that addresses this soon.

goldyfruit:
@stmcginnis There is no rush, until then I'll use Ansible to collect these information.
Thanks!

PR #52, opened 2019-12-31
Add access to Bios Attributes property

stmcginnis:
Attributes are a collection of key/value pairs that vary based in implementation. This adds a type and helper methods for the most common value types. The attributes can still be accessed as a map, with direct handling needed for the type of the value.

Closes: #49

### DIFF
--- a/redfish/bios.go
+++ b/redfish/bios.go
@@ -7,12 +7,51 @@ package redfish
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/stmcginnis/gofish/common"
 )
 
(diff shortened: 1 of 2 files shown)


stmcginnis:
OK, finally had a chance to test this a little more against a real system and it appears to be working as expected. Not sure I really like this solution, but since I haven't come up with a better alternative yet, let's go with this for now and see if there are any issues or better ideas raised.

```go
	headers = []string{"Key", "Value"}
	systems, _ := service.Systems()
	for _, system := range systems {
		bios, _ := system.Bios()
		for key := range bios.Attributes {
			data = append(data, []string{
				key,
				bios.Attributes.String(key),
			})
		}
	}
```

PR #53, opened 2020-01-29
Support changing TLS handshake timeout

treydock:


treydock:
Only finding this necessary on Dell systems with IDRACs that around 4 years old and somewhat slow to respond to redfish.

stmcginnis:
Looks like something that could be useful. Thanks!

PR #56, opened 2020-02-19
send reset action if server return empty supported reset type

leslie-qiwa:
If server returns empty supported reset type, it should be reset without any reset type. Seems like this is the behavior of hpe server

### DIFF
--- a/redfish/manager.go
+++ b/redfish/manager.go
@@ -358,18 +358,25 @@ func ListReferencedManagers(c common.Client, link string) ([]*Manager, error) {
 
 // Reset shall perform a reset of the manager.
 func (manager *Manager) Reset(resetType ResetType) error {
+	if len(manager.SupportedResetTypes) == 0 {
+		// reset directly without reset type. HPE server has the behavior
+		type temp struct {
+			Action string
+		}
+		t := temp{
+			Action: "Manager.Reset",
+		}
+
+		_, err := manager.Client.Post(manager.resetTarget, t)
+		return err
+	}
 	// Make sure the requested reset type is supported by the manager.
 	valid := false
-	if len(manager.SupportedResetTypes) > 0 {
-		for _, allowed := range manager.SupportedResetTypes {
-			if resetType == allowed {
-				valid = true
-				break
-			}
+	for _, allowed := range manager.SupportedResetTypes {
+		if resetType == allowed {
+			valid = true
+			break
 		}
-	} else {
-		// No allowed values supplied, assume we are OK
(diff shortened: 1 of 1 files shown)


stmcginnis:
This worries me a little. What if the client requests something like GracefulShutdown. In this case, we would just ignore what they asked for and reset the server instead of shutting it down. Probably forcefully, depending on how HPE has implemented that.

Is this against an older version of ILO? It might help to understand what version it is and whether they have an updated firmware that changes this behavior.

redfish/manager.go
```diff
@@ -358,18 +358,25 @@ func ListReferencedManagers(c common.Client, link string) ([]*Manager, error) {
...
-				break
-			}
+	for _, allowed := range manager.SupportedResetTypes {
+		if resetType == allowed {
+			valid = true
+			break
 		}
-	} else {
-		// No allowed values supplied, assume we are OK
-		valid = true
```
stmcginnis: If we do this, I think it would be better to just change the action here rather than moving the check out separately.

leslie-qiwa: I thought that too at beginning, but seems like the structure need change from 
```
	type temp struct {
		ResetType ResetType
	}
```
to 
```
	type temp struct {
		Action string
	}
```

leslie-qiwa:
> This worries me a little. What if the client requests something like GracefulShutdown. In this case, we would just ignore what they asked for and reset the server instead of shutting it down. Probably forcefully, depending on how HPE has implemented that.
> 
> Is this against an older version of ILO? It might help to understand what version it is and whether they have an updated firmware that changes this behavior.

This is `iLO 5 1.45 Jun 27 2019`. 

Totally understand your concern. Another approach is to create a new function for that. What do you think?

leslie-qiwa:
Seems like `iLO 2.10 Oct 30 2019` does return two Types. I'll explore more about it. Thanks for the comment.

stmcginnis:
I suppose is if this is present from versions from mid-2019, it's going to be more prevalent than if it was just an old outdated version.

It would be good if there was logging in place so we could at least log a warning if this is the case so there's a better chance for the consumer to realize what is going on. But since this seems to be a bug in iLO, I guess I'm fine working around it and being able to point to that implementation as the reason something like a GracefulShutdown didn't happen like expected.

So looks fine to me. I would hope most systems will not hit this code, but at least there's the code comment as a breadcrumb in case anyone ever needs to dig into why it's behaving the way it is.

Thanks for fixing this and looking into it!

PR #57, opened 2020-02-19
add GracefulRestart reset type

leslie-qiwa:
HPE iLO support this reset type

### DIFF
--- a/redfish/computersystem.go
+++ b/redfish/computersystem.go
@@ -350,14 +350,20 @@ type ResetType string
 const (
 	// OnResetType shall be used to power on the machine
 	OnResetType ResetType = "On"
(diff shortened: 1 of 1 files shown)


stmcginnis:
Looks like we're also missing PowerCycle and PushPowerButton. Want to add those too while you're at it?

http://redfish.dmtf.org/schemas/v1/Resource.json

leslie-qiwa:
> Looks like we're also missing PowerCycle and PushPowerButton. Want to add those too while you're at it?
> 
> http://redfish.dmtf.org/schemas/v1/Resource.json

Good suggestion. Done.

stmcginnis:
Thanks @leslie-qiwa!

PR #58, opened 2020-03-10
Fixed AccountService Links field

danjpar:
Redfish AccountService output does not have Links field. Both Accounts and Roles are contained in the main body.
Would you prefer to keep it this way to prevent confusion with the Accounts() and Roles() methods?

### DIFF
--- a/examples/change_login.go
+++ b/examples/change_login.go
@@ -1,0 +1,52 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+package main
+
+import (
+	"fmt"
+
+	"github.com/stmcginnis/gofish"
+)
+
+func main() {
+	// Create a new instance of gofish client, ignoring self-signed certs
+	username := "my-username"
+	config := gofish.ClientConfig{
+		Endpoint: "https://bmc-ip",
+		Username: username,
+		Password: "my-password",
+		Insecure: true,
+	}
+	c, err := gofish.Connect(config)
+	if err != nil {
+		panic(err)
+	}
+	defer c.Logout()
+
+	// Retrieve the service root
+	service := c.Service
+
+	// Query the chassis data using the session token
+	accts, err := service.AccountService()
+	if err != nil {
+		panic(err)
+	}
+	accs, err := accts.Accounts()
+	if err != nil {
+		panic(err)
+	}
+
+	for _, acc := range accs {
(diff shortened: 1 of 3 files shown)


danjpar:
I will correct+add comments to the example file tomorrow

examples/change_login.go
```diff
@@ -0,0 +1,53 @@
...
+		panic(err)
+	}
+	accs, err := accts.Accounts()
+	if err != nil {
+		panic(err)
+	}
+
+	for _, acc := range accs {
+		if acc.UserName == username {
+			urlE := "/redfish/v1/AccountService/Accounts/"+acc.ID
```
stmcginnis: We shouldn't ever expose a redfish URL to the consumer of the library.

stmcginnis:
Thanks @danjpar, this looks like a good start. We need to encapsulate the Redfish internals though.

stmcginnis:
So what I think we need to do is in the Account structure, capture and store the account's URI in the Get of the object:

https://github.com/stmcginnis/gofish/blob/e2c7648f228a4fb6b7e3fa67a53098ee07c8d966/redfish/accounts.go#L121

Then have an UpdateAccount() call or similar to set new values that uses that stored URI to know where to make the call to do the update.

What I had actually been planning on was a more general Save() or Update() mechanism for all objects. I think what probably needs to be done is to have some shadow values stored of an objects original values. Then on the Update() call, use a stored URI that is part of common.Entity and figure out which values have been updated to send the changes to the server.

danjpar:
We can add a URI field with json:"@odata.id" to unmarshal, then add save and update methods to accept a payload and either do a put post or patch depending

stmcginnis:
Awesome, let's go with this for now. Thanks!

danjpar:
Should I also add ODataContext, ODataEtag, and ODataType to common.Entity then remove from all structs that have those fields and an Entity reference?

Searching for common.Entity, 5 structs also have ODataID but none of the others: powercontrol, redundancy, storagecontroller, fan, temperature

accountservice, account, role, chassis, powersupply, and voltage have no OData fields

stmcginnis:
Thanks, but I don't see any need for those currently. I would rather wait until there is a need so that we are not consuming memory we don't need to be.

If I remember the spec right (I need to go back and look, probably this weekend), not all of the objects have an ODataID.

PR #60, opened 2020-04-22
Added support of basic authentication

mikeletux:
Hello Sean,

I've added the possibility of basic authentication against the redfish API. This might be interesting in different scenarios (like the ones for some reason you don't want to have a token).

To the gofish.ClientConfig struct I've added a field called BasicAuth. If not set, it will use the token authentication as before. If set explicitly to true, it will use basic auth. 

To allow this to happen, I've also modified the struct redfish.AuthToken, adding Username, Password and a BasicAuth fields. This will make us be aware if we want to use token auth or basic auth while building the APIClient.

When creating the request for HTTP methods, we will check this struct and we will include the needed headers, either for token auth or basic auth.

Best regards.
/Miguel

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ package gofish
 import (
 	"bytes"
 	"crypto/tls"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


stmcginnis:
Thanks Miguel! I don't have something to test this against, but I assume you have. We can always fix any issues with follow ups.

Thanks for the contribution!

PR #69, opened 2020-06-04
Add virtual media and publish raw data

LRichi:
1. Add VirtualMedia for management virtual media
2. Publish rawData for extends custom features any vendors


.idea/gofish.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stmcginnis, line 1: And remove these idea files.

redfish/virtual_media.go
```diff
@@ -0,0 +1,81 @@
...
+	NotConnectedVirtualMediaConnectedMethod VirtualMediaConnectedMethod = "NotConnected"
+	AppletVirtualMediaConnectedMethod       VirtualMediaConnectedMethod = "Applet"
+	OemVirtualMediaConnectedMethod          VirtualMediaConnectedMethod = "Oem"
+)
+
+// VirtualMedia is used to represent virtual media resource.
+type VirtualMedia struct {
+	common.Entity
+
+	ODataContext        string                      `json:"@odata.context"` // ODataContext is the odata context.
```
stmcginnis, line 33: godoc comments should be above the line they document.

redfish/virtual_media.go
```diff
@@ -0,0 +1,81 @@
+//
```
stmcginnis, line 1: Please rename file to virtualmedia.go to match the object name.

stmcginnis: If you run tools/gen_schema.sh, that will generate this file under gofiles and will automatically pull in the spec descriptions as docs for each public struct and field.

redfish/virtual_media.go
```diff
@@ -0,0 +1,81 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package redfish
+
+import (
+	"encoding/json"
+	"github.com/LRichi/gofish/common"
```
stmcginnis, line 9: This will need to be updated to the right namespace.

LRichi:
I made changes

PR #71, opened 2020-06-09
Handle HREF in links but prefer @odata.id

galthaus:
For really old HP/HPE systems, they do not use @odata.id, but use href instead.  This allows the system to continue to limp along and work better.  At least get past session creation.

You may not want this, but it seems to work on my old HP system and continues working with my dell system.

Clicked too quickly on some things.  Edited to add more information.


stmcginnis:
Thanks @galthaus. I think this is a great way that the library can hide some of these incompatibilities so the consumer of the code doesn't need to even know about it. Looks good!

PR #73, opened 2020-06-24
Etag changed from string to json.Number

mikeletux:
Hello Sean,

Looks like the *ODataEtag* field from *AccountService* struct was set to **string**. Apparently this is what Redfish spec says.
However, I was testing this against my infra (Dell R7415 with iDRAC9 version 4.00.00) and the body from the response was:

~~~
{
	"@odata.context": "/redfish/v1/$metadata#AccountService.AccountService",
	"@odata.type": "#AccountService.v1_3_1.AccountService",
	"@odata.id": "/redfish/v1/AccountService",
	"@odata.etag": 77968587947,
	"AccountLockoutCounterResetAfter": 0,
	"AccountLockoutDuration": 0,
	"AccountLockoutThreshold": 0,
	"Accounts": {
		"@odata.id": "/redfish/v1/AccountService/Accounts"
	},
...
}
~~~

As you can see, in this case (for some reason I truly unknown) the returned JSON comes with that field set to an **int**. 
I set that field to json.Number to avoid that situation, since it supports froat64, int64 and string. That solves the issue.

At the same time I will try to contact PG from PowerEdge to see what's going on in there.

This issue might affect other parts of the code when dealing with PowerEdge probably... If I find more stuff, I will report it in the future :)

### DIFF
--- a/redfish/accountservice.go
+++ b/redfish/accountservice.go
@@ -159,7 +159,7 @@ type AccountService struct {
 	// ODataContext is the odata context.
 	ODataContext string `json:"@odata.context"`
 	// ODataEtag is the odata etag.
-	ODataEtag string `json:"@odata.etag"`
+	ODataEtag json.Number `json:"@odata.etag"`
 	// ODataType is the odata type.
 	ODataType string `json:"@odata.type"`
 	// AccountLockoutCounterResetAfter shall contain the


stmcginnis:
Hmm, unfortunately I think we may need to get this fixed in iDRAC rather than here. ETags should be strings, so if we made this change it would likely break others.

We may need to see if there is some way we can handle this in unmarshalljson for that object that would be able to detect that it is being passed as a number, then do the string comparison there.

Another option that comes to mind, and maybe it's the easiest at this point since we are not actually using this Etag info anywhere in the gofish code, would be to just drop that field completely.

mikeletux:
Hi Sean,

json.Number is not gonna break things. Behind the scenes is a string but comes with methods to convert from int or float to string. Take a look (from [link](https://golang.org/src/encoding/json/decode.go?s=7244:7286#L184))

~~~
// A Number represents a JSON number literal.

type Number string


// String returns the literal text of the number.

func (n Number) String() string { return string(n) }


// Float64 returns the number as a float64.

func (n Number) Float64() (float64, error) {

	return strconv.ParseFloat(string(n), 64)

}


// Int64 returns the number as an int64.

func (n Number) Int64() (int64, error) {

	return strconv.ParseInt(string(n), 10, 64)

}
~~~

But yeah, I agree this also needs to be fixed from PowerEdge side. Just by changing that gives a little more resilience to the code since other vendors might have mess up with that as well.

Also, if you prefer to delete it... Personally I'm not going to use it, so up to you :)

Best regards.
/Miguel

stmcginnis:
True, it won't break things as far as holding the value that is returned. But as it is now, it would expose up a json data type that any consumer would need to use.

Maybe we can go somewhere inbetween. In unmarshalljson we could have a temporary field that would convert the returned value using json.Number, then set the real object's field using the String() call.

But I am kind of leaning towards just dropping all Etags until we actually see a need for them.

mikeletux:
Yes, please, then let's do that 👍

PR #74, opened 2020-06-24
Remove etag properties

stmcginnis:
ETags are defined as strings, but some vendor implementations return this as a numeric value, causing problems in the JSON unmarshalling from the API response. Since these are not used anywhere in the library, and since it is mostly an internal protocol detail, this just removes the field for now until we know we actually need it.


stmcginnis:
@mikeletux, would be great to know if this works for you.

PR #77, opened 2020-07-09
Fix Logout fail

sungup:
I propose this PR to fix Logout() fail on some vendor and have a question.

According to Redfish specification, client must get the URI of Session Resource from the Location header. But some vendor implementation returns the full URI path like `https://blahblah.blah/redfish/v1/SessionService/Sessions/{session id}`. This URI will generate the invalid session address and fail the logout from BMC, because last runRequest() from the call-path of Logout() combines that full URI path with the endpoint URL. If a user run this as a client tool many times, it makes session overflow.

From that reason I modified the CreateSession, but I didn't put any code to return an error from Logout() on failure.

Is it necessary to return an error when a logout fails?

### DIFF
--- a/redfish/session.go
+++ b/redfish/session.go
@@ -6,6 +6,7 @@ package redfish
 
 import (
 	"encoding/json"
+	"net/url"
 
 	"github.com/stmcginnis/gofish/common"
 )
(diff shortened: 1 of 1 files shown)


stmcginnis:
"But some vendor implementation returns the full URI path" sounds like an implementation bug that the vendor needs to fix. That said, we can certainly try to handle it better in gofish.

I would be concerned with how this would work when behind a NAT or proxy where the host name the Redfish service knows itself as doesn't match what the client uses to access the host. I wonder if it would be better to look for and strip the leading part of the URL (anything before "redfish/v1") to make sure what the client uses does not get lost.

Just some initial thoughts. I will try to take a closer look at this and try it out soon. Thanks!

stmcginnis:
OK, I got a chance to take a closer look and now I see what you're doing. Good idea!

I tested this out against some other systems and looks good. At least not any worse than it was before. ;)

PR #78, opened 2020-07-29
Adding ability to manage the redfish event subscriptions.

vinialves:
### Adding ability to manage to manage the event subscriptions using the event service.

The idea is to expose four new functions in the event service:

**To create an event subscription:**
```
// CreateEventSubscription creates the subscription using the event service.
// destination should contain the URL of the destination for events to be sent.
// eventTypes is a list of EventType to subscribe to.
// httpHeaders is optional and gives the opportunity to specify any arbitrary
// HTTP headers required for the event POST operation.
// oem is optional and gives the opportunity to specify any OEM specific properties,
// it should contain the vendor specific struct that goes inside the Oem session.
// It returns the new subscription URI if the event subscription is created
// with success or any error encountered.
func (eventservice *EventService) CreateEventSubscription(
	destination string,
	eventTypes []EventType,
	httpHeaders map[string]string,
	oem interface{},
) (string, error) {

```

**To delete an event subscription:**
```
// DeleteEventSubscription deletes a specific subscription using the event service.
func (eventservice *EventService) DeleteEventSubscription(uri string) error {

```

**To get all event subscriptions:**
```
// GetEventSubscriptions gets all the subscriptions using the event service.
func (eventservice *EventService) GetEventSubscriptions() ([]*EventDestination, error) {
```

**To get a specific event subscription:**
```
// GetEventSubscription gets a specific subscription using the event service.
func (eventservice *EventService) GetEventSubscription(uri string) (*EventDestination, error) {
```

Also, I am adding the new type EventType to be used in some places:
- EventService should contain a list of EventType as EventTypesForSubscription.
- EventDestination should contain a list of EventType as EventTypes.
- The CreateEventSubscription also receives a list list of EventType to subscribe to.

Added validation for the input parameters for all these new functions, also added unit tests for all new functions.

In order to create tests for the new functions I am also improving the test client to give the possibility to define custom returns for the actions.

Also, I am adding unit test for the recent issue about Logout "some vendor implementation returns the Location header with full URI address" as I also found this issue before it was fixed. The subscription URI returned by CreateEventSubscription is also taking in consideration this possible situation.

Also, I am changing some hardcoded constants, like **"POST"**, to use the values defined in http package, for example: **http.MethodPost**.

I have performed some tests with real Redfish servers and all new functions worked as expected. Thanks for the nice work, gofish is working great for me. Let me know your thoughts about these suggestions of changes/improvements. Thanks!

### DIFF
--- a/client.go
+++ b/client.go
@@ -160,27 +160,27 @@ func (c *APIClient) Get(url string) (*http.Response, error) {
 		relativePath = common.DefaultServiceRoot
 	}
 
-	return c.runRequest("GET", relativePath, nil)
+	return c.runRequest(http.MethodGet, relativePath, nil)
 }
 
 // Post performs a Post request against the Redfish service.
 func (c *APIClient) Post(url string, payload interface{}) (*http.Response, error) {
-	return c.runRequest("POST", url, payload)
+	return c.runRequest(http.MethodPost, url, payload)
 }
 
 // Put performs a Put request against the Redfish service.
 func (c *APIClient) Put(url string, payload interface{}) (*http.Response, error) {
-	return c.runRequest("PUT", url, payload)
+	return c.runRequest(http.MethodPut, url, payload)
 }
 
 // Patch performs a Patch request against the Redfish service.
 func (c *APIClient) Patch(url string, payload interface{}) (*http.Response, error) {
-	return c.runRequest("PATCH", url, payload)
+	return c.runRequest(http.MethodPatch, url, payload)
 }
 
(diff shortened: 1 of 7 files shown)


stmcginnis:
Great updates @vinialves - thanks! This looks good to me. I haven't had a chance to test on my systems yet, but everything looks in order and I can always follow up if there are actually any issues found once I do. Sounds like you tested well, and I really appreciate the improvements to unit testing.

Issue #79, opened 2020-08-06
Add session management


mbethke:
I'm currently developing a Redfish-based Icinga check and would like to avoid spamming my iLO event log with "admin connection" entries - and I think this would go towards avoiding problems like in #75, too. The idea is to get a session token once and preserve it in-between plugin invocations by writing it to a session file, performing as new login only if the old one should have timed out. To do this, I'd need `APIClient.auth` exported though. At least that's how I hacked it for now - although that way I have to basically duplicate the second part of `gofish.Connect()`; perhaps it would be better to add an Auth member to `ClientConfig` and check in `Connect()` whether this or a username has been provided.

stmcginnis:
Sounds reasonable. You mention having a hack in place to do this. Feel free to propose a PR with changes, and we can work through whether it fits or not. Thanks!

mbethke:
Sorry, forgot to close this when you merged #83. All good now :)

PR #82, opened 2020-08-21
export symbol client.Auth

mbethke:
This is the hack I mentioned in #79. It works fine but required a bit of code duplication regarding session initialization. I use it like this (slightly simplified):
```golang
// This initializes a new client with either an existing or a new session
func client(endpoint, user, pass string) (client *gofish.APIClient, err error) {
        client, err = oldSession(endpoint, "mysessionfile")
        if err != nil {
                return newSession(endpoint, sessionfile, user, pass)
        }
        return client, nil
}

func oldSession(endpoint, sessionfile string) (client *gofish.APIClient, err error) {
        tok, err := readSession(sessionfile) // returns a redfish.AuthToken
        if err != nil {
                return
        }
        client, err = connect(endpoint, "", "")
        if err != nil {
                return
        }
        service, err := gofish.ServiceRoot(client)
        if err != nil {
                return
        }

        client.Service = service
        client.Auth = tok
        _, err = service.Chassis()
        // Tried a query; if it errors out, this session has probably expired
        if err != nil {
                return
        }
        return client, nil
}

func newSession(endpoint, sessionfile, user, pass string) (client *gofish.APIClient, err error) {
        client, err = connect(endpoint, user, pass)
        if err != nil {
                return
        }
        err = writeSession(sessionfile, client.Auth)
        return
}

func connect(endpoint, user, pass string) (c *gofish.APIClient, err error) {
        return gofish.Connect(
                gofish.ClientConfig{
                        Endpoint:   endpoint,
                        Username:   user,
                        Password:   pass,
                        Insecure:   true,
                })
}
```
I'll just open another PR with an idea how to avoid this.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,7 +38,7 @@ type APIClient struct {
 	Service *Service
 
 	// Auth information saved for later to be able to log out
-	auth *redfish.AuthToken
+	Auth *redfish.AuthToken
 
 	// dumpWriter will receive HTTP dumps if non-nil.
 	dumpWriter io.Writer
@@ -128,7 +128,7 @@ func Connect(config ClientConfig) (c *APIClient, err error) {
 		}
 
 		client.Service = service
-		client.auth = auth
+		client.Auth = auth
 	}
 
 	return client, err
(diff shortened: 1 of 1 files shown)


mbethke:
Thanks for merging #83! :+1:
I'll close this one as obsolete.

PR #83, opened 2020-08-21
Add session data to ClientConfig

mbethke:
This adds a `Session` field to `ClientConfig`, to pass in session data from a previous run, and `GetSession()` to retrieve it from a client for serialization. Smells slightly cleaner to me, but it's totally untested except that it compiles. If you think this looks like a better approach than simply exporting `Auth`, I'd change my client to use it like this and give it some testing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,6 +44,13 @@ type APIClient struct {
 	dumpWriter io.Writer
 }
 
+// Session holds the session ID and auth token needed to identify an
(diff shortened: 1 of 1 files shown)


stmcginnis:
Sorry for the long delay. Somehow I missed this notification.

This looks reasonable to me. Since it is optional, should be safe to add it. Thanks!

Issue #85, opened 2020-09-10
APIClient.Delete method just return error and not the HTTP response itself


mikeletux:
Hello Sean,

I've noticed that the *APIClient.Delete() method just returns an error interface saying if there was any issue with that call. The rest of HTTP method calls return also the a *http.Response object.
I think having also the *http.Response object returned from the Delete method is also very useful. Let me highlight that with an example. Please see the image below:

![deletemethod](https://user-images.githubusercontent.com/6768988/92742692-2af0a980-f380-11ea-8fd4-550a21845ab5.PNG)

I'm trying to get rid of a virtual disk. To do that, I need to perform a Delete onto /redfish/v1/Systems/\<ComputerSystem-Id>/Storage/\<Storage-Id>/Volumes.  

The operation is not done rightaway. In my case, the iDRAC, which is the controller that implements the RedfishAPI, creates a job. Then I need to record the job completion status, to see if that was ok or not.
In the Delete response headers, you can see the **Location** header, which is a link to that. I need to get that for checking the job status. Unfortunately in the current implementation I cannot get that because the Delete() method only returns an error.

Would you be open to accept a PR where apart from the error status also returns a *http.Response object?

Many thanks!
/Miguel

stmcginnis:
I think that's a good idea @mikeletux. It's a change that may break some existing code, but I think you are correct that the current implementation is not right, and I think that should be changed to be consistent with the other calls.

mikeletux:
Issue closed in #86

PR #86, opened 2020-09-11
Add *http.Response return to APIClient.Delete method

mikeletux:
Hello Sean,

I'm proposing the PR in reference to [!82](https://github.com/stmcginnis/gofish/issues/85).

I've little tested the new APIClient.Delete call and it works. I also needed to change some other files that used that call. I've always followed the philosophy of being as less intrusive in the code as possible. 

Now the Client interface looks like this (please check the **Delete** definition):
~~~
type Client interface {
	Get(url string) (*http.Response, error)
	Post(url string, payload interface{}) (*http.Response, error)
	Patch(url string, payload interface{}) (*http.Response, error)
	Put(url string, payload interface{}) (*http.Response, error)
	Delete(url string) (*http.Response, error)
}
~~~

Files also changed apart from client.go and types.go:
- eventdestination.go
- session.go

Thanks!
/Miguel

### DIFF
--- a/client.go
+++ b/client.go
@@ -179,15 +179,15 @@ func (c *APIClient) Patch(url string, payload interface{}) (*http.Response, erro
 }
 
 // Delete performs a Delete request against the Redfish service.
-func (c *APIClient) Delete(url string) error {
+func (c *APIClient) Delete(url string) (*http.Response, error) {
 	resp, err := c.runRequest(http.MethodDelete, url, nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	if resp != nil && resp.Body != nil {
 		resp.Body.Close()
 	}
-	return nil
+	return resp, nil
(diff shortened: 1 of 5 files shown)


mikeletux:
As far as I know the CI pipeline fails because of the tests not being able to use the new signature.
Should we modify the Mocked client and rest of tests?

stmcginnis:
Yeah, I think things should pass once you update the mocked client. That needs to track the real client.

mikeletux:
yeah, I think so 😄 
I will push an update, hopefully this weekend solving that

mikeletux:
Hello Sean,

I've done the changes needed to the mock client as well as the needed dependencies.

Let me know.
Best regards.
/Miguel

stmcginnis:
Thanks @mikeletux, looks great!

PR #87, opened 2020-09-22
Volume OperationApplyTimeSupport added

mikeletux:
Hello Sean,

I've created this pull request to address something that was missing in the current implementation.

I needed to figure out the *OperationApplyTimeValues* for Storage controllers. This is very useful to see if you can perform a disk operation rightraway (Immediate) or you need to reboot the server (OnReset).

Having these values is highly valuable to see what strategy to follow when it comes to create virtual disks on top of disk controllers 😄 

Let me know!
/Miguel

### DIFF
--- a/redfish/volume.go
+++ b/redfish/volume.go
@@ -192,15 +192,6 @@ type Volume struct {
 	drives []string
 }
 
-// Volumes is used to represent the volumes information related to a Storage
-type Volumes struct {
-	common.Entity
-	// OperationApplyTimeSupport contains, among other things, the types
-	// of apply times the client is allowed request when performing a Create,
-	// Delete, or Action operation.
-	OperationApplyTimeSupport common.OperationApplyTimeSupport `json:"@Redfish.OperationApplyTimeSupport"`
-}
-
 // UnmarshalJSON unmarshals a Volume object from the raw JSON.
 func (volume *Volume) UnmarshalJSON(b []byte) error {
 	type temp Volume
@@ -290,15 +281,17 @@ func AllowedVolumesUpdateApplyTimes(c common.Client, link string) ([]common.Oper
 		return nil, err
 	}
 	defer resp.Body.Close()
-	var volumes Volumes
+	var temp struct {
+		OperationApplyTimeSupport common.OperationApplyTimeSupport `json:"@Redfish.OperationApplyTimeSupport"`
+	}
 
-	err = json.NewDecoder(resp.Body).Decode(&volumes)
+	err = json.NewDecoder(resp.Body).Decode(&temp)
 	if err != nil {
 		return nil, err
 	}
 
 	var applyTimes []common.OperationApplyTime
-	for _, v := range volumes.OperationApplyTimeSupport.SupportedValues {
+	for _, v := range temp.OperationApplyTimeSupport.SupportedValues {
 		applyTimes = append(applyTimes, v)
 	}
 	return applyTimes, nil
--- a/redfish/volume.go
+++ b/redfish/volume.go
@@ -192,6 +192,15 @@ type Volume struct {
 	drives []string
 }
 
+// Volumes is used to represent the volumes information related to a Storage
+type Volumes struct {
+	common.Entity
+	// OperationApplyTimeSupport contains, among other things, the types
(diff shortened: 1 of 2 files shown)


redfish/volume.go
```diff
@@ -192,6 +192,15 @@ type Volume struct {
 	drives []string
 }
 
+// Volumes is used to represent the volumes information related to a Storage
```
stmcginnis, line 195: The storage.volumes property is actually a VolumesCollection. That collection is access via the existing GetReferencedVolumes call below.

The existing volume struct should be used rather creating a duplicate here. See bios.go for an existing example of extracting ApplyTimes so we can stay consistent with how this is done.

stmcginnis (commented):
Thanks for adding this. Some minor changes needed I think, so we can keep the API consistent.

redfish/volume.go
mikeletux: Hi @stmcginnis ,

I think it is a little bit different from the bios case. In the Bios case we have an struct where we unmarshal the response from the redfish API. Calling to */redfish/v1/Systems/System.Embedded.1/Bios* gives you, among other things, the *@Redfish.Settings* for that.

In the case of Volumes, the endpoint is */redfish/v1/Systems/System.Embedded.1/Storage/RAID.Integrated.1-1/Volumes* and produces a result like this one:
~~~
{
	"@Redfish.OperationApplyTimeSupport": {
		"@odata.type": "#Settings.v1_2_1.OperationApplyTimeSupport",
		"SupportedValues": [
			"Immediate",
			"OnReset"
		]
	},
	"@odata.context": "/redfish/v1/$metadata#VolumeCollection.VolumeCollection",
	"@odata.id": "/redfish/v1/Systems/System.Embedded.1/Storage/RAID.Integrated.1-1/Volumes",
	"@odata.type": "#VolumeCollection.VolumeCollection",
	"Description": "Collection Of Volume",
	"Members": [
		{
			"@odata.id": "/redfish/v1/Systems/System.Embedded.1/Storage/RAID.Integrated.1-1/Volumes/Disk.Virtual.0:RAID.Integrated.1-1"
		}
	],
	"Members@odata.count": 1,
	"Name": "Volume Collection"
}
~~~

It gives you the *Redfish.OperationApplyTimeSupport*. That value (to my understanding) gives you the supported operations on that disk controller. I need those values.
Take into account that I'm verifying that on the Volumes endpoint before creating a volume (Volumes collection is empty at that point).

Using the current Volume structure would be like querying */redfish/v1/Systems/System.Embedded.1/Storage/RAID.Integrated.1-1/Volumes/Disk.Virtual.0:RAID.Integrated.1-1*, but I don't want to do that.

How could I possibly get *@Redfish.OperationApplyTimeSupport* on */redfish/v1/Systems/System.Embedded.1/Storage/WHATEVER_CONTROLLER/Volumes* endpoint as per the current implementation?

Thanks!
/Miguel

mikeletux:
Hello Sean,

I think I got it. I removed the Volumes structure, because as you said, it is a Collection.
Created an additional aux struct within the AllowedVolumesUpdateApplyTimes func to grab the fields I needed.
Hope it is compliant with the rest of the API.

Regards,
/Miguel

stmcginnis:
Thanks Miguel, I do think this fits better with the rest of the behavior of the client.

I do wonder if it would be better to update the storage.Volumes() call to return a tuple of the []Volume's and the ApplyUpdateTimes. Maybe something to think about. The good thing is, I think having both would be fine, so there's no reason to  not move ahead with this.

One thing I would like to do as a follow up, especially since you have captured the response JSON, would be to add that to the unit test and assert that the parsing works as expected. But again, not anything to hold this up. I will do that as a follow up to this change later, unless you feel like taking care of it in another PR. Either way is perfectly fine with me.

mikeletux:
Hello Sean,

Thanks for merge that into the code 😄 
Actually I'm going back and forth with gofish client and the terraform provider for redfish and unfortunately I don't have that much time to write the tests. If you have some time to do that, I'd appreciate it. Otherwise I'll try to do that in my free time.

Best regards!

stmcginnis:
No problem at all Miguel. Thanks for your contribution!

Issue #89, opened 2020-10-20
Access links object for manager, ethernet interfaces, etc


leslie-qiwa:
Need access links object to further retrieve more information. For example, need to retrieve dell OEM attributes through manager API. 
```
  "Links": {
    "Oem": {
      "Dell": {
        "DellAttributes": [
          {
            "@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/Attributes"
          },
          {
            "@odata.id": "/redfish/v1/Managers/System.Embedded.1/Attributes"
          },
          {
            "@odata.id": "/redfish/v1/Managers/LifecycleController.Embedded.1/Attributes"
          }
        ],
```

stmcginnis:
Closed by #90

PR #90, opened 2020-10-20
expose manager's Links object to allow caller to further process

leslie-qiwa:

It is for https://github.com/stmcginnis/gofish/issues/89.

### DIFF
--- a/redfish/manager.go
+++ b/redfish/manager.go
@@ -217,6 +217,10 @@ type Manager struct {
 	// ManagerNetworkProtocol which represents the network services for this
 	// manager.
 	networkProtocol string
+	// OEMData are all OEM data under top level manager section
+	OEMData map[string]interface{}
+	// OEMLinks are all OEM data under link section
+	OEMLinks map[string]interface{}
 	// PartNumber shall contain a part number assigned by the organization that
 	// is responsible for producing or manufacturing the manager.
 	PartNumber string
@@ -304,6 +308,7 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 		ManagerForSwitches      common.Links
 		ManagerForSwitchesCount int `json:"ManagerForSwitches@odata.count"`
 		ManagerInChassis        common.Link
+		OEM                     map[string]interface{} `json:"Oem"`
 	}
 	var t struct {
 		temp
@@ -315,6 +320,7 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 		VirtualMedia         common.Link
 		Links                linkReference
 		Actions              actions
+		OEM                  map[string]interface{} `json:"Oem"`
 	}
 
 	err := json.Unmarshal(b, &t)
@@ -327,6 +333,8 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 	manager.ethernetInterfaces = string(t.EthernetInterfaces)
 	manager.logServices = string(t.LogServices)
 	manager.networkProtocol = string(t.NetworkProtocol)
+	manager.OEMData = t.OEM
+	manager.OEMLinks = t.Links.OEM
 	manager.remoteAccountService = string(t.RemoteAccountService)
 	manager.serialInterfaces = string(t.SerialInterfaces)
 	manager.virtualMedia = string(t.VirtualMedia)
--- a/common/types.go
+++ b/common/types.go
@@ -60,7 +60,16 @@ func (e *Entity) Update(originalEntity reflect.Value, currentEntity reflect.Valu
 		fieldName := originalEntity.Type().Field(i).Name
 		originalValue := originalEntity.Field(i).Interface()
 		currentValue := currentEntity.Field(i).Interface()
-		if originalValue != currentValue {
+		if originalValue == nil && currentValue == nil {
+			continue
+		} else if originalValue == nil {
+			payload[fieldName] = currentValue
+		} else if reflect.TypeOf(originalValue).Kind() != reflect.Map {
+			if originalValue != currentValue {
+				// TODO: Handle JSON name being different than field name
+				payload[fieldName] = currentValue
+			}
+		} else if !reflect.DeepEqual(originalValue, currentValue) {
 			// TODO: Handle JSON name being different than field name
 			payload[fieldName] = currentValue
 		}
--- a/redfish/manager_test.go
+++ b/redfish/manager_test.go
@@ -6,12 +6,46 @@ package redfish
 
 import (
 	"encoding/json"
+	"github.com/stmcginnis/gofish/common"
+	"reflect"
 	"strings"
 	"testing"
-
-	"github.com/stmcginnis/gofish/common"
 )
 
+var oemLinksBody = `
+			{
+				"Dell": {
+					"DellAttributes": [
+						{
+							"@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/Attributes"
+						},
+						{
+							"@odata.id": "/redfish/v1/Managers/System.Embedded.1/Attributes"
+						},
+						{
+							"@odata.id": "/redfish/v1/Managers/LifecycleController.Embedded.1/Attributes"
+						}
+					],
+					"DellAttributes@odata.count": 3,
+					"DellTimeService": {
+						"@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/DellTimeService"
+					}
+				}
+			}
+`
+var oemDataBody = `
+		{
+			"Dell": {
+				"DelliDRACCard": {
+					"@odata.context": "/redfish/v1/$metadata#DelliDRACCard.DelliDRACCard",
+					"@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/DelliDRACCard/iDRAC.Embedded.1-1_0x23_IDRACinfo",
+					"@odata.type": "#DelliDRACCard.v1_1_0.DelliDRACCard",
+					"IPMIVersion": "2.0",
+					"URLString": "https://10.5.1.83:443"
+				}
+			}
+		}
+`
 var managerBody = `{
 		"@Redfish.Copyright": "Copyright 2014-2019 DMTF. All rights reserved.",
 		"@odata.context": "/redfish/v1/$metadata#Manager.Manager",
@@ -92,8 +126,10 @@ var managerBody = `{
 			],
 			"ManagerInChassis": {
 				"@odata.id": "/redfish/v1/Chassis/Chassis-1"
-			}
-		},
+			},
+			"Oem":
+` + oemLinksBody +
+`		},
 		"Actions": {
 			"#Manager.Reset": {
 				"target": "/redfish/v1/Managers/BMC-1/Actions/Manager.Reset",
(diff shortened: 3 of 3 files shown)


redfish/manager.go
```diff
@@ -322,9 +324,19 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
...
+	}
+	err = json.Unmarshal(b, &allLinks)
+	if err != nil {
+		return err
+	}
+
 	// Extract the links to other entities
 	*manager = Manager(t.temp)
 	manager.ethernetInterfaces = string(t.EthernetInterfaces)
+	manager.Links = allLinks.Links
```
stmcginnis, line 339: Links are already parsed below. Those are needed for getting the links to other resources.

This doesn't look like what you were looking for in the issue filed though. Based on that, you want to get links that are returned under the OEM section. That's a bit different than here. I haven't thought of a good, intuitive way to get OEM data since it could include links, strings, or other nested objects depending on each individual vendor.

So I don't believe this approach is right, but unfortunately I don't have a good answer for that right now.

leslie-qiwa: yes, I noticed it has parsed with explicit fields, but I have to double parse because of current implementation. Agree it is not elegant.

I'm thinking another two possible solutions:
 - solution 1: add LinkOEM field whose type is map[string]interface{}, to expose this section. This will be same as other field, and can embed into `type linkReference struct {` for parsing.
 - solution 2: have one Raw method to cache and return original whole section, and let caller parse if current data is not enough. This seems more generic. 

What do you think?

stmcginnis: I did play around with having an OEM field of map[string]interface{} once, but I seem to recall there being some issue with it with different returned OEM data. If you get something like that working, I think that would be great. We just need to make sure to try it against a few different vendor implementations to make sure it actually covers the different options.

I don't think LinkOEM is the right name for it though. Probably need an OEMData or OEMField(key) method seems like a better fit. Some of the returned values may be links, but most OEM data I've seen is actually string or integer values.

leslie-qiwa: The reason I proposes `LinkOEM` is due to Dell Manager body has 2 OEM sections at different places, such as below partial data. Let me know what would be the best name in this case? Once we had the name, I'll test the patch against both HPE and DELL.

```
{
  "@odata.context": "/redfish/v1/$metadata#Manager.Manager",
  "@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1",
  "@odata.type": "#Manager.v1_6_0.Manager",
  "Id": "iDRAC.Embedded.1",
  "Links": {
    "Oem": {
      "Dell": {
        "DellAttributes": [
          {
            "@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/Attributes"
          },
          {
            "@odata.id": "/redfish/v1/Managers/System.Embedded.1/Attributes"
          },
          {
            "@odata.id": "/redfish/v1/Managers/LifecycleController.Embedded.1/Attributes"
          }
        ],
        "DellAttributes@odata.count": 3,
        "DellJobService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellJobService"
        },
        "DellLCService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellLCService"
        },
        "DellLicensableDeviceCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellLicensableDeviceCollection"
        },
        "DellLicenseCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellLicenseCollection"
        },
        "DellLicenseManagementService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellLicenseManagementService"
        },
        "DellOpaqueManagementDataCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellOpaqueManagementDataCollection"
        },
        "DellPersistentStorageService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellPersistentStorageService"
        },
        "DellSwitchConnectionCollection": {
          "@odata.id": "/redfish/v1/Dell/Systems/System.Embedded.1/NetworkPorts/DellSwitchConnectionCollection"
        },
        "DellSwitchConnectionService": {
          "@odata.id": "/redfish/v1/Dell/Systems/System.Embedded.1/DellSwitchConnectionService"
        },
        "DellSystemManagementService": {
          "@odata.id": "/redfish/v1/Dell/Systems/System.Embedded.1/DellSystemManagementService"
        },
        "DellSystemQuickSyncCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellSystemQuickSyncCollection"
        },
        "DellTimeService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellTimeService"
        },
        "DellUSBDeviceCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellUSBDeviceCollection"
        },
        "DelliDRACCardService": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DelliDRACCardService"
        },
        "DellvFlashCollection": {
          "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellvFlashCollection"
        },
        "Jobs": {
          "@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/Jobs"
        }
      }
    }
  },
  "Oem": {
    "Dell": {
      "DelliDRACCard": {
        "@odata.context": "/redfish/v1/$metadata#DelliDRACCard.DelliDRACCard",
        "@odata.id": "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DelliDRACCard/iDRAC.Embedded.1-1_0x23_IDRACinfo",
        "@odata.type": "#DelliDRACCard.v1_1_0.DelliDRACCard",
        "IPMIVersion": "2.0",
        "LastSystemInventoryTime": "2020-10-19T19:32:53+00:00",
        "LastUpdateTime": "2020-10-19T21:43:41+00:00",
        "URLString": "https://10.5.1.83:443"
      }
    }
  },
  "PowerState": "On",
  "Status": {
    "Health": "OK",
    "State": "Enabled"
  },
  "UUID": "3330484f-c0c8-3180-5410-00544c4c4544"
}
```

stmcginnis: Oh, good point. I was thinking of the latter case, but you are correct - there is Manager.Oem and Manager.Links.Oem. So extracting the OEM links out to something accessible should be good. I would suggest OEMLinks rather than LinksOEM.

leslie-qiwa: Sounds good. I will make two fields: `OEMLinks` and `OEMData`

leslie-qiwa:
@stmcginnis Please review.

- Use `interface{}` as type to make it more generic. 
- Also seems like both `OEM` and `Oem` are used in the project. In order to be align with json parsing, I used `OemData` and `OemLinks`. What do you think?

leslie-qiwa:
tested on dell/hpe/cisco, and get expected result.

redfish/manager.go
```diff
@@ -217,6 +217,10 @@ type Manager struct {
 	// ManagerNetworkProtocol which represents the network services for this
 	// manager.
 	networkProtocol string
+	// OEMData are all OEM data under top level manager section
+	OemData interface{}
+	// OEMLinks are all OEM data under link section
+	OemLinks interface{}
```
stmcginnis, line 223: Please use common.Links

redfish/manager.go
```diff
@@ -304,6 +308,7 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 		ManagerForSwitches      common.Links
 		ManagerForSwitchesCount int `json:"ManagerForSwitches@odata.count"`
 		ManagerInChassis        common.Link
+		Oem                     interface{}
```
stmcginnis, line 311: To keep with Go style, name these OEM and use `json:"Oem"` to match the JSON name (though IIRC, the json serializer will handle that).

redfish/manager.go
```diff
@@ -315,6 +320,7 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 		VirtualMedia         common.Link
 		Links                linkReference
 		Actions              actions
+		Oem                  interface{}
```
stmcginnis, line 323: I think common.Links should be used here.

redfish/manager.go
stmcginnis: I think for OEM properties, we can likely use something better than interface{}. For OEM properties, the specification defines them as having a name as a key and a JSON object as a property. So I think we should at least provide that list of keys with interface{} values so the consumer can decide what to do once they identify the value they want.

stmcginnis (commented):
Thanks. Please also add unit test coverage.

redfish/manager.go
leslie-qiwa: common.Links seems can not pass UT if data like below
```
			"Oem": {
				"Dell": {
					"DellAttributes": [
						{
							"@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/Attributes"
						},
						{
							"@odata.id": "/redfish/v1/Managers/System.Embedded.1/Attributes"
						},
						{
							"@odata.id": "/redfish/v1/Managers/LifecycleController.Embedded.1/Attributes"
						}
					],
					"DellAttributes@odata.count": 3,
					"DellTimeService": {
						"@odata.id": "/redfish/v1/Managers/iDRAC.Embedded.1/DellTimeService"
					}
				}
			}
```
Below is UT failure. 
```
manager_test.go:189: Error decoding JSON: json: cannot unmarshal object into Go struct field linkReference.Oem of type common.Links
```

redfish/manager.go
leslie-qiwa: I like map[string]interface too, but seems like it will have crash in common.(*Entity).Update because it tries to compare map and decide if it need update. I can even change `Update` method if you like, but feel it is more than this PR.
```
panic: runtime error: comparing uncomparable type map[string]interface {} [recovered]
	panic: runtime error: comparing uncomparable type map[string]interface {}

goroutine 18 [running]:
testing.tRunner.func1(0xc000124100)
	/usr/local/go/src/testing/testing.go:830 +0x392
panic(0x1316f60, 0xc0000ad8a0)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/stmcginnis/gofish/common.(*Entity).Update(0xc00009ca80, 0x1380de0, 0xc00009d500, 0x199, 0x1380de0, 0xc00009ca80, 0x199, 0xc000149ee8, 0x3, 0x3, ...)
	/Users/qiwa/workspace/golang/src/github.com/stmcginnis/gofish/common/types.go:63 +0x3b8
github.com/stmcginnis/gofish/redfish.(*Manager).Update(0xc00009ca80, 0xc0000b1080, 0xc00009ca80)
```

```
		fieldName := originalEntity.Type().Field(i).Name
		originalValue := originalEntity.Field(i).Interface()
		currentValue := currentEntity.Field(i).Interface()
		if originalValue != currentValue {
			// TODO: Handle JSON name being different than field name
			payload[fieldName] = currentValue
		}
```

redfish/manager.go
leslie-qiwa: I changed to map[string]interface{}. Seems like common.Links have issues to parse sample objects.

stmcginnis:
Thanks for the updates @leslie-qiwa. This looks good now. I may make a few tweaks once I get some time to play around with this, but no need to hold things up until then.

Thanks!

PR #91, opened 2020-10-20
add ID of network adapter

leslie-qiwa:
align with https://redfish.dmtf.org/schemas/v1/NetworkAdapter.v1_5_0.json

Close https://github.com/stmcginnis/gofish/issues/92

### DIFF
--- a/redfish/networkadapter.go
+++ b/redfish/networkadapter.go
@@ -181,6 +181,8 @@ type NetworkAdapter struct {
 	Controllers []Controllers
 	// Description provides a description of this resource.
 	Description string
+	// ID is the identifier of the network adapter.
+	ID string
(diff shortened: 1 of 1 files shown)


stmcginnis:
Good catch, it's even one of the required properties:

            "required": [
                "@odata.id",
                "@odata.type",
                "Id",
                "Name"
            ],

Can you add it to the unit test data?

leslie-qiwa:
seems like it is already in this common entity. I need close it
```
type Entity struct {
	// ODataID is the location of the resource.
	ODataID string `json:"@odata.id"`
	// ID uniquely identifies the resource.
	ID string `json:"Id"`
	// Name is the name of the resource or array element.
	Name string `json:"Name"`
	// Client is the REST client interface to the system.
	Client Client
}
```

PR #93, opened 2020-10-21
Added UpdateService to the API

mikeletux:
Hello Sean,

Hope you're doing well 😄 

Recently I've noticed that the gofish API was missing from the *UpdateService* part, which allow operators to update the firmware from their infrastructure.
What I've done is to implement the standard way of upgrading firmware (via **/redfish/v1/UpdateService/Actions/UpdateService.SimpleUpdate**) and left apart the "OEM" functions (which vary from vendor to vendor, i.e. in Dell infra would be **/redfish/v1/UpdateService/Actions/Oem/DellUpdateService.Install**).
This should be portable across vendors.

Also I've added the unit tests needed for the new struct (*UpdateService*) and JSON unmarshalling.

Best regards!

### DIFF
--- a/redfish/updateservice.go
+++ b/redfish/updateservice.go
@@ -1,0 +1,78 @@
+package redfish
+
+import (
+	"encoding/json"
+
+	"github.com/stmcginnis/gofish/common"
+)
+
+// UpdateService is used to represent the update service offered by the redfish API
+type UpdateService struct {
+	common.Entity
+
+	// ODataContext is the odata context.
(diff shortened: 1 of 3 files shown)


stmcginnis:
Thanks!

Just a note that I will be adding the license headers to the two files you added. Ideally, that should be done by the contributor to be explicit about the granted license. I am assuming that's OK, but just let me know if there are any concerns.

mikeletux:
@stmcginnis yeah, no worries from my side. Go ahead and add the current license to those files.

Issue #94, opened 2020-10-29
Error structure in 400 reply


leslie-qiwa:
When the request could not be processed because it contains missing or invalid information (such as validation error on an input field, a missing required value, and so on), redfish will return an extended error in the response body, as defined in clause Error Responses. To help client make smart decision, it is better that the package can detect and return the error if possible.

refer 
http://redfish.dmtf.org/schemas/DSP0266_1.0.html#status-codes
http://redfish.dmtf.org/schemas/DSP0266_1.0.html#error-responses

PR #95, opened 2020-10-30
return formated error struct for status code 400

leslie-qiwa:
Signed-off-by: Leslie Wang <qiwa@pensando.io>

Close https://github.com/stmcginnis/gofish/issues/94


stmcginnis:
Awesome, thanks for the great test coverage. This looks good to me.

PR #97, opened 2020-11-13
add protocol and context arguments for event subscriptions

dhellmann:
I had to add Protocol and Context to the payload so Dell would accept
the event subscription definition.

I did that by adding extra arguments all the way through the stack. Since
they are required, I added them before the optional `oem` argument to
those functions.


stmcginnis:
Looks good - thanks Doug!

PR #105, opened 2020-12-15
Add ability to upgrade client security to session-based and do multi-part POST requests

DanDanN00dle:
Some actions require a session token to perform. Rather than require client users to remember credentials and create new clients add a method to clone an existing client and upgrade the security using the stored credentials.

Additionally, some actions (e.g. firmware upload) require a multi-part POST of the firmware to the server. To this end this provides a new function `PostMultipart` that takes and forms a multi-part payload and POSTs it to the requested URL.

Finally, to aide in troubleshooting allow dynamic setting of the client `DumpWriter` to allow targeted request/response tracing and add a `go.mod` file so programs using the library can use a local version easier.

### DIFF
--- a/client.go
+++ b/client.go
@@ -171,6 +171,31 @@ func ConnectDefault(endpoint string) (c *APIClient, err error) {
 	return client, err
 }
 
+// CloneWithSession will create a new Client with a session instead of basic auth.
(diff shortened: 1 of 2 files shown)


stmcginnis:
Looks good to me. If you have time, it would be nice to have an example of using this code. That can be a follow up though.

PR #106, opened 2020-12-16
Implement software inventory objects and link off Update service for Firmware Inventory

DanDanN00dle:
This change modifies the `UpdateService` to change the `FirmwareInventory` string (URL) to a method to fetch the corresponding `SoftwareInventoryCollection`. From that collection hangs the individual `SoftwareInventory` entries (`Members()`) that correspond to items in the `FirmwareInventory`.

Also provided an example (`upload_firmware.go`) for how to query the firmware inventory of a server and push a firmware update using a multipart POST request.

### DIFF
--- a/examples/upload_firmware.go
+++ b/examples/upload_firmware.go
@@ -1,0 +1,106 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/stmcginnis/gofish"
+)
+
+func main() {
+	// Create a new instance of gofish client, ignoring self-signed certs
+	config := gofish.ClientConfig{
+		Endpoint: "https://bmc-ip",
+		Username: "my-username",
+		Password: "my-password",
+		Insecure: true,
+	}
+
+	c, err := gofish.Connect(config)
+	if err != nil {
+		panic(err)
+	}
+	defer c.Logout()
+
+	// Attached the client to service root
+	service := c.Service
+
+	// Query the update service
+	update, err := service.UpdateService()
+	if err != nil {
+		panic(err)
+	}
+
+	// Fetch the firmware inventory
+	inventory, err := update.FirmwareInventory()
+	if err != nil {
+		panic(err)
+	}
+
+	// Fetch the inventory members
+	firmwares, err := inventory.Members()
+	if err != nil {
+		panic(err)
+	}
+
+	// Loop over the inventory, printing the details
+	for _, f := range firmwares {
+		fmt.Printf("%v@%v: %v", f.Name, f.Version, f.Description)
+	}
+
+	// Get the session token
+	session, err := c.GetSession()
+	if err != nil {
+		// Didn't have a token. Upgrade to a session
+		c, err = c.CloneWithSession()
+		if err != nil {
+			panic(err)
+		}
+		defer c.Logout()
+	}
+
+	// Open the firmware binary
+	firmwareReader, err := os.Open("firmware.bin")
+	if err != nil {
+		panic(err)
+	}
+	defer firmwareReader.Close()
+
+	if service.Vendor == "HPE" {
+		// For HPE systems, need a "parameters" JSON blob
+		parameters := map[string]interface{}{
+			"UpdateRepository": true,
+			"UpdateTarget":     true,
+			"ETag":             "sometag",
+			"Section":          0,
+		}
+
+		// Marshal that to a reader
+		parameterBytes, err := json.Marshal(parameters)
+		if err != nil {
+			panic(err)
+		}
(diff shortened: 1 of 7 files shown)


examples/upload_firmware.go
```diff
@@ -0,0 +1,106 @@
+package main
```
stmcginnis, line 1: If you do an update, would you mind throwing an SPDX header in this file?

redfish/softwareinventory.go
```diff
@@ -0,0 +1,62 @@
...
+	Status                 common.Status
+	Version                string
+	LowestSupportedVersion string
+	Manufacturer           string
+	ReleaseDate            string
+	SoftwareID             string
+	UefiDevicePaths        []string
+	Updateable             bool
+	WriteProtected         bool
+	rawData                []byte
```
stmcginnis, line 27: Since this object doesn't support setting values via the Update() call, the capture of the rawData could be removed.

redfish/softwareinventory.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	err := json.Unmarshal(b, &t)
+	if err != nil {
+		return err
+	}
+
+	// Extract the links to other entities for later
+	*softwareInventory = SoftwareInventory(t.temp)
+	softwareInventory.rawData = b
```
stmcginnis, line 44: Then if we don't need to capture rawData, the whole UnmarshallJSON call can be removed too and just go with the default unmarshalling.

redfish/softwareinventorycollection.go
```diff
@@ -0,0 +1,70 @@
+//
```
stmcginnis, line 1: This file should be dropped (see note in updateservice.go).

redfish/softwareinventorycollection_test.go
```diff
@@ -0,0 +1,112 @@
+//
```
stmcginnis, line 1: This file should be dropped (see note in updateservice.go).

redfish/updateservice.go
```diff
@@ -58,13 +58,18 @@ func (updateService *UpdateService) UnmarshalJSON(b []byte) error {
...
-	updateService.FirmwareInventory = string(t.FirmwareInventory)
+	updateService.firmwareInventory = string(t.FirmwareInventory)
 	updateService.TransferProtocol = t.Actions.SimpleUpdate.AllowableValues
 	updateService.UpdateServiceTarget = t.Actions.SimpleUpdate.Target
 	updateService.rawData = b
 	return nil
 }
 
+// FirmwareInventory will get the firmware inventory from the service
+func (updateService *UpdateService) FirmwareInventory() (*SoftwareInventoryCollection, error) {
```
stmcginnis, line 69: This will make this call inconsistent with the pattern followed elsewhere in the library. Everywhere else where there is a collection of related objects, the library takes care of that internal collection mechanism and just returns the desired objects to the users.

An example of this pattern can be seen here:

https://github.com/stmcginnis/gofish/blob/df4eea3f71d6c35a6dc7094ea936e2fd4d4f2091/redfish/chassis.go#L409

The collections object is consistent across the various \*Collection objects, so we can treat them all the same. So this should be updated to return the set of FirmwareInventory objects. And firmwareinventory.go will just need to add the ListReferenced\* convention used with the other objects to extract each instance out of the collections, like what is done here:

https://github.com/stmcginnis/gofish/blob/df4eea3f71d6c35a6dc7094ea936e2fd4d4f2091/redfish/networkadapter.go#L258

stmcginnis (changes requested):
Nice! Thanks for working on this. A few issues pointed out inline to keep this consistent with the rest of the library, but this looks great.

redfish/updateservice.go
vinialves: I agree with the changes requested by @stmcginnis to follow the library pattern.

Also, note that Redfish UpdateService "/redfish/v1/UpdateService/" has the links for "FirmwareInventory" and "SoftwareInventory":

```
{
    "@odata.context": "/redfish/v1/$metadata#UpdateService.UpdateService",
    "@odata.etag": "W/\"97FD6D18\"",
    "@odata.id": "/redfish/v1/UpdateService/",
    "@odata.type": "#UpdateService.v1_1_1.UpdateService",
    "Id": "UpdateService",

	...
	
	"FirmwareInventory": {
        "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/"
    }
	
	... 
	
	"SoftwareInventory": {
        "@odata.id": "/redfish/v1/UpdateService/SoftwareInventory/"
    }
}

```

Looking at Redfish schema, my understanding is that both links in UpdateService should return a list of SoftwareInventory.

vinialves: I have created a new PR to better explain the point here: https://github.com/stmcginnis/gofish/pull/123

PR #107, opened 2021-01-05
Initial work to handle Redfish registries.

vinialves:
This is a proposal of initial work to handle Redfish registries.

It adds 5 new functions in service root:

1) Registries gets the Redfish Registries
2) MessageRegistries gets all the available message registries in all languages
3) MessageRegistriesByLanguage gets the message registries by language.
4) MessageRegistryByLanguage gets a specific message registry by language.
5) MessageByLanguage tries to find and get the message in the correct language from the informed messageID.

These new functions can be useful in a scenario where an application creates event subscriptions and receive events where in the event we have the messageId and the application should discover the details about the event message, like the message, number of args, and others from the available Redfish Registries considering the desired language (useful for internationalization of applications). More information can be found here: https://www.dmtf.org/sites/default/files/Redfish%20School%20-%20Events.pdf

I was also thinking about add another function similar to the new function MessageByLanguage  that also would receive the message arguments and return a final processed message, but after some consideration, I think It should not be a responsability for the library.

Created unit test files for all new files.

Performed some tests with all new functions using real Redfish server and Gofish returned the correct objects/results.

Please, let me know if it looks good and if you have any suggestions of improvements.

### DIFF
--- a/redfish/messageregistry.go
+++ b/redfish/messageregistry.go
@@ -1,0 +1,257 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package redfish
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/stmcginnis/gofish/common"
+)
+
+// MessageRegistryMessage is a message contained in the message registry.
+type MessageRegistryMessage struct {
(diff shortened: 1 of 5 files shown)


stmcginnis:
Looks great, thanks @vinialves.

Issue #104, opened 2020-12-14
pass context throughout


jacobweinstock:
Hey, great library! thanks so much for all the hard work!

I was wondering if you'd be open to a PR to add context throughout? It'd obviously be a pretty breaking change as most of the function/method signatures would need to be changed.

stmcginnis:
Thanks Jacob. Could you explain a little more what you are trying to do?

jacobweinstock:
Ah sorry, yes. I'm interested in being able to pass in a context and have `gofish` honor the timeout, deadline, or cancelation.  

Right now, I know I can set a timeout by using the timeout in the `http.Client` struct and passing that into `Connect` or using the `ClientConfig`.  My challenge is that when importing `gofish` and using it with other libraries they can pass around a context and use/understand the timeout/deadline/cancel functionality. 

With current `gofish` I need to translate the context timeout/deadline to a number and then use that in `gofish`. I also haven't found a clean way to honor a canceled context. 

Hope that makes sense.

stmcginnis:
I do think that would end up being too disruptive.

Really, the client object is the context that already gets passed around. Can we add something there that would help you?

jacobweinstock:
Sure, totally understand. The challenge with the client object is that putting `context.Context` into a struct is not ideal or recommended. 
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.
https://golang.org/pkg/context/

Thanks for the conversation and the consideration though. I will have to rethink on my end how to proceed.

vinialves:
Hello guys, good discussion here!

Connect receives the struct ClientConfig, and as explained the recommended golang way to send the context is sending it as the first parameter in the function args.

To keep Connect working for previous Gofish versions and to avoid sending Context in the ClientConfig struct (that is not the prefered idiom to send the context as explained) we could create a func ConnectContext(ctx context.Context, config ClientConfig) (c *APIClient, err error)

We could change Connect and ConnectDefault to get the context from context.Background()

In order to implement as recommended by golang we would need to do changes in all functions to send the context as first parameter to "runRequest" in "client.go" where we would replace the NewRequest call:
req, err := http.NewRequest(method, endpoint, payloadBuffer)
req, err := http.NewRequestWithContext(ctx, endpoint, payloadBuffer)

[...]

PR #108, opened 2021-01-05
Changes to send the context to the HTTP request.

vinialves:
This proposal adds 2 new functions to do the connection with context:

ConnectContext - The same as Connect, but sets the ctx.
ConnectDefaultContext - The same as ConnectDefault, but sets the ctx.

Because we already have some authentication code inside Connect, the new ConnectContext is not able to reuse Connect with the correct context. So, actually it was required some changes in Connect and ConnectDefault to avoid too much code duplication. 

The context is defined in the client struct to avoid a lot of function/method signature modifications. It is not the preferred Golang idiom to send the context through the functions, but it is still ok in my opinion and it seems other libs do the same approach (probably to avoid to many modifications as would be required here). This could be an initial approach to send the context and maybe in the future we could rework to remove the context from the client struct.

Also, changed the way the request is created in order to pass the context, starting using http.NewRequestWithContext.

Performed some tests with ConnectContext with real Redfish server and Gofish returned the correct error when the context timeout. Also, tested the normal Connect function with real Redfish server and it kept working as expected.

Added unit tests to cover context timeout and context cancellation.

Please, let me know if it looks good and if you have any suggestions of improvements.

Related discussion can be found here: https://github.com/stmcginnis/gofish/issues/104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: go
 
 go:
-- 1.12.x
+- 1.15.x
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package gofish
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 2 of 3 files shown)


stmcginnis:
This looks very reasonable to me. Thanks for the good test coverage and validating against a real system. Let's go with this and see if we get any feedback from others.

PR #110, opened 2021-01-13
Adding a more simple and performatic function to get a redfish message registry.

vinialves:
The search for the Redfish message registry by registry/language can be very slow depending on the environment (existing number of registries / size of the registries). 

Created new function to get the MessageRegistry directly from an uri, avoiding doing unecessary searches to find it.


stmcginnis:
Thanks!

PR #113, opened 2021-01-15
Adding oem to redfish update service.

vinialves:
Adding oem to redfish update service.

Some changes to get oem for update service as specified in Redfish.
Also, improving the existing unit test to cover the change.

Thanks!


stmcginnis:
I had hoped to find a good way to provide consistent access to OEM properties, but it looks like the definition is flexible enough that we can't even just use map[string]interface{}. So I think this is reasonable to just make it interface{} and leave it to the consumer to figure out what the property actually contains.

Thanks for the test coverage!

PR #114, opened 2021-01-19
Return Error struct also for 404

mikeletux:
Hi Sean,

Working in our Terraform project, I found very useful that when the Redfish API returns also a 404, the client returns an Error struct rather than a generic error. That will allow us to actually distinguish between an actual error or a simple "doesn't exist".

I'm sending this PR to meet this specific case. I need to find out if the GetVolume() function below actually fails or returns 404.

~~~
v, err := redfish.GetVolume(c, "/redfish/v1/Systems/System.Embedded.1/Storage/RAID.Integrated.1-1/Volumes/Disk.Virtual.0:RAID.Integrated.1-1")
	if val, ok := err.(*common.Error); !ok {
		panic(err)
	}
~~~

Please let me know.
Best regards!
/Miguel

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -725,7 +725,7 @@ func ConstructError(statusCode int, b []byte) error {
 type Error struct {
 	rawData []byte
 	// An integer that represents the status code returned by the API
-	HTTPReturnedStatusCode int
+	HTTPReturnedStatusCode int `json:"-"`
 	// A string indicating a specific MessageId from the message registry.
 	Code string `json:"code"`
 	// A human readable error message corresponding to the message in the message registry.
--- a/common/types.go
+++ b/common/types.go
@@ -716,13 +716,16 @@ func ConstructError(statusCode int, b []byte) error {
 		// return normal error
 		return fmt.Errorf("%d: %s", statusCode, string(b))
 	}
+	err.Error.HTTPReturnedStatusCode = statusCode
 	err.Error.rawData = b
 	return err.Error
 }
 
-// Error is redfish error response object for 400 or 404
+// Error is redfish error response object for HTTP status codes different from 200, 201 and 204
 type Error struct {
 	rawData []byte
+	// An integer that represents the status code returned by the API
+	HTTPReturnedStatusCode int
 	// A string indicating a specific MessageId from the message registry.
 	Code string `json:"code"`
 	// A human readable error message corresponding to the message in the message registry.
--- a/common/types.go
+++ b/common/types.go
@@ -712,7 +712,7 @@ func ConstructError(statusCode int, b []byte) error {
 	var err struct {
 		Error *Error
 	}
-	if e := json.Unmarshal(b, &err); e != nil || err.Error == nil{
+	if e := json.Unmarshal(b, &err); e != nil || err.Error == nil {
 		// return normal error
 		return fmt.Errorf("%d: %s", statusCode, string(b))
 	}
@@ -720,7 +720,7 @@ func ConstructError(statusCode int, b []byte) error {
 	return err.Error
 }
 
-// Error is redfish error response object for 400
+// Error is redfish error response object for 400 or 404
 type Error struct {
 	rawData []byte
 	// A string indicating a specific MessageId from the message registry.
--- a/client.go
+++ b/client.go
@@ -440,11 +440,7 @@ func (c *APIClient) runRawRequest(method string, url string, payloadBuffer io.Re
 			return nil, err
 		}
 		defer resp.Body.Close()
-		if resp.StatusCode == 400 || resp.StatusCode == 404 {
-			return nil, common.ConstructError(resp.StatusCode, payload)
-		}
-
-		return nil, fmt.Errorf("%d: %s", resp.StatusCode, string(payload))
+		return nil, common.ConstructError(resp.StatusCode, payload)
 	}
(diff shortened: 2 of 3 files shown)


mikeletux:
I saw the build failing because of test *func TestErrorNon400(t \*testing.T)*

What is the reason why that struct is only being returned for 400 errors?

/Miguel

stmcginnis:
I'm not sure if the spec calls for an extended error to be returned on a 404. You will have to check the JSON response from the system to verify that, and likely handle the case if it doesn't return an extended error in case other implementations don't. I'd have to reread that part of the spec to be sure.

The reason it fails is because you have not updated the unit tests for the client code. If you capture the returned JSON, it should be easy to update that test code and use the JSON data as the data to be parsed.

mikeletux:
As per my implementation on DellEMC infra, yes, it returns extended info. But no worries, for me it is ok to keep it like it is right now. My problem is another one...

The only thing I'm struggling with, is to figure out if an 404 response is returned outside the scope of the client. At the moment  the following checks are done:
~~~
if resp.StatusCode != 200 && resp.StatusCode != 201 && resp.StatusCode != 202 && resp.StatusCode != 204
~~~ 
If that is met, a response is returned. If other return code is returned, nil alongside the error is returned. Since error is very generic for my usecase... Would you be open to return the response even though the response doesn't meet those three statusCodes ?

Thanks!
/Miguel

stmcginnis:
The resp.Body is returned in the error. Is there something missing from that that you need?

Though looking at ConstructError, it should be fine to use that no matter what the status code is. It will try to extract the Redfish Error struct, and if it can't, it basically just falls back to formatting the string like it is conditionally in the call in the client.

https://github.com/stmcginnis/gofish/blob/20eeed20881f18a25040e33911e8335e0dc8ba78/common/types.go#L711-L721

So I think best bet is probably to just swap out the few places in the client code where there is this conditional check and always have it call ConstructError.

mikeletux:
Yeah, the solution you propose sounds great. If response is different from 200, 201 or 204, return the struct no matter what.
For error checking, you can try to extract it, and if it fails it is a generic error. If it doesn't, you can check what happened in the call.

Also, would you agree to add a field in Error struct called *HTTPReturnedStatusCode int* where we set what happened? 

Thanks!
/Miguel

mikeletux:
I've updated the code for the *Error* struct as well as the code from the *runRawRequest()*.

Please let me know what you think about this so I can update the code that tests that.

Thanks!
/Miguel

mikeletux:
@stmcginnis did you have time to take a look into this? 😄

common/types.go
```diff
@@ -712,17 +712,20 @@ func ConstructError(statusCode int, b []byte) error {
 	var err struct {
 		Error *Error
 	}
-	if e := json.Unmarshal(b, &err); e != nil || err.Error == nil{
+	if e := json.Unmarshal(b, &err); e != nil || err.Error == nil {
 		// return normal error
 		return fmt.Errorf("%d: %s", statusCode, string(b))
```
stmcginnis, line 717: I wonder if it might be better to always return the Error object. We could handle this unmarshalling error by manually setting err.Error.Message = string(b). That way any consuming client wouldn't need to have conditional checking on what was returned. It can always be assumed to be a Redfish Error object, even when another error has occurred. Then they can always access the error message the same way.

common/types.go
```diff
@@ -712,17 +712,20 @@ func ConstructError(statusCode int, b []byte) error {
...
 	err.Error.rawData = b
 	return err.Error
 }
 
-// Error is redfish error response object for 400
+// Error is redfish error response object for HTTP status codes different from 200, 201 and 204
 type Error struct {
 	rawData []byte
+	// An integer that represents the status code returned by the API
+	HTTPReturnedStatusCode int
```
stmcginnis, line 728: Seems reasonable. And the JSON marshalling will safely ignore this field since the return data won't include it.

stmcginnis (commented):
@mikeletux - looks like a reasonable approach overall. Some thoughts inline.

Still need to update the unit tests to pass.

@leslie-qiwa - if you have a minute, would be great if you could chime in on this since you originally added this bit. Also wondering if the rawData is being used for something. Can that be removed?

leslie-qiwa:
@stmcginnis @mikeletux I like the idea adding `HTTPReturnedStatusCode int` because my original struggle is to get status code. `ConstructError:Error()` only returns redfish error body, but not status code. This new change is a better solution to expose the information as well as simplifying the codes. :-)

mikeletux:
Awesome guys 😄 
I will push some test for that during this weekend, so we can merge it asap!

Thanks!
/Miguel

mikeletux:
Hey @stmcginnis @leslie-qiwa ,

I've push the changes already. Got rid of **TestErrorNon400** since now we are capturing all non 200, 201 and 204. Also included a check within **TestError400** that actually checks if the return code is 400:
~~~
if errStruct.HTTPReturnedStatusCode != 400 {
	t.Errorf("The error code is different from 400")
}
~~~

It would be great if you see it ok and could merge it 😄 
Have a great weekend!
/Miguel

stmcginnis:
Awesome, thanks. This looks great. I will probably push a few minor follow up tweaks, but I think we can get this updated and I will tag a new release.

PR #118, opened 2021-01-25
TestClient updated according APIClient

mikeletux:
Hi Sean,

Just pushing this PR to update **TestClient** struct to be as similar as possible to **APIClient** struct (considering recent changes on **APIClient** regarding the returned Error struct).

Let me know 😄 
Best regards!

/Miguel


stmcginnis:
Good call!

PR #122, opened 2021-02-04
Added VirtualMediaConfig struct and new method

mikeletux:
Hello Sean,

Sorry I had to open another pull request. I've reset the two commits I had in my repo that implemented the old solution.

As we agreed, I've created a new exportable (public) struct called *VirtualMediaConfig* that can be passed to the new method *InsertMediaConfig*. Unfortunately Go doesn't support method overloading so another name had to be used. I think that method name should be ok. Tests were implemented as well.

Also I've kept what I did in the previous implementation regarding EjectMedia and avoiding sending no body at all, just sending an empty JSON object to avoid issues with different Redfish implementations.

Let me know 😄 
Best regards!

/Miguel

### DIFF
--- a/redfish/virtualmedia.go
+++ b/redfish/virtualmedia.go
@@ -200,7 +200,7 @@ func (virtualmedia *VirtualMedia) EjectMedia() error {
 		return errors.New("redfish service does not support VirtualMedia.EjectMedia calls")
 	}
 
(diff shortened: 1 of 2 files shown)


stmcginnis:
Works for me, thanks Miguel!

PR #123, opened 2021-02-09
Adding ability to list firmware and software inventories from Redfish UpdateService.

vinialves:
This PR adds ability to list firmware and software inventories from Redfish updateService.
More details here: https://github.com/stmcginnis/gofish/pull/106/files#r573121773

@stmcginnis  I noted there is a script to help to generate the objects from the Redfish schema "gen_script.sh", very nice. 
This time I used the script to generate softwareinventory.go. Just cleared a little bit the generated file to have only some basic initial functionalities. Also, changed somethings in the generated file, for example the Oem type to be an interface{} We can fix this in the script later.

I copied the test file "softwareinventory_test.go" from @DanDanN00dle PR (https://github.com/stmcginnis/gofish/pull/106) and the tests are passing.

Also, I did a test with a real Redfish server and I was able to list firmware and software inventories from using updateService.

Some suggestions: 
This PR can be used just for reference to do the additional implementations requested in https://github.com/stmcginnis/gofish/pull/106
Or we can merge this PR and close https://github.com/stmcginnis/gofish/pull/106 then if @DanDanN00dle wants he could create a new PR to merge the additional code (the upload firmware example application, etc).

Thanks!

### DIFF
--- a/redfish/softwareinventory.go
+++ b/redfish/softwareinventory.go
@@ -1,0 +1,104 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package redfish
+
+import (
+	"encoding/json"
+
+	"github.com/stmcginnis/gofish/common"
+)
+
+// SoftwareInventory is This Resource contains a single software
+// component that this Redfish Service manages.
+type SoftwareInventory struct {
+	common.Entity
+
+	// ODataContext is the odata context.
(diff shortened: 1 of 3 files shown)


stmcginnis:
Awesome, thanks @vinialves. Let's wait a bit and see if there is any response with the other PR. If not, let's go ahead with this and Dan can follow up like you suggest.

PR #125, opened 2021-02-24
Expose ManagedBy property  in ComputerSystem

iurygregory:
The ManagedBy property  of ComputerSystem is not exposed
in the main struct, the property  is available since v1.0.0
see [1]

[1] http://redfish.dmtf.org/schemas/v1/ComputerSystem.1.0.0.json


stmcginnis:
Thanks!

PR #126, opened 2021-03-16
Fix ManagedBy in ComputerSystem

iurygregory:
If Marshal and Unmarshal are executed in a CS struct
with ManagedBy set, the struct after Unmarshal will
have an empty ManagedBy list. This commit fix this
problem and add tests to cover this scenario.

### DIFF
--- a/redfish/computersystem_test.go
+++ b/redfish/computersystem_test.go
@@ -248,6 +248,12 @@ func TestComputerSystem(t *testing.T) {
 		t.Errorf("Invalid allowable reset actions, expected 6, got %d",
 			len(result.SupportedResetTypes))
 	}
+	if len(result.ManagedBy) != 1 {
+		t.Errorf("Received invalid number of ManagedBy: %d", len(result.ManagedBy))
+	}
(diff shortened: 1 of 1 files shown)


redfish/computersystem.go
```diff
@@ -568,7 +568,6 @@ func (computersystem *ComputerSystem) UnmarshalJSON(b []byte) error {
 	computersystem.resetTarget = t.Actions.ComputerSystemReset.Target
 	computersystem.SupportedResetTypes = t.Actions.ComputerSystemReset.AllowedResetTypes
 	computersystem.setDefaultBootOrderTarget = t.Actions.SetDefaultBootOrder.Target
-	computersystem.ManagedBy = t.Links.ManagedBy.ToStrings()
```
stmcginnis, line 571: This just removes ManagedBy from ever being set. Was there somewhere else you were thinking of handling this?

iurygregory: Yeah I just noticed, I'm trying to figure out why the value set in `ManagedBy`  is available in the new struct after doin Marshal and UnMarshal =(

stmcginnis (approved):
Thanks!

PR #133, opened 2021-04-05
Client interface and TestClient updated

mikeletux:
Hi Sean,

In my previous PR I forgot to update the Client interface as well as the TestClient to accomodate new methods (XXXWithHeaders as well as Multipart ones).

Maybe in the future, before 1.0 release we can re-design the client interface as well as the APIClient to just have another argument for customHeaders in get/post/put/patch/delete😄 .

Let me know!
/Miguel

### DIFF
--- a/common/testclient.go
+++ b/common/testclient.go
@@ -7,6 +7,7 @@ package common
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Yeah, good idea to clean up before a 1.0 release. Thanks for taking care of this.

Issue #147, opened 2021-05-19
Drive identifiers sometimes not an array


sosodev:
I noticed that `chassis.Drives()` was often returning a truncated slice of drives. A little bit of debugging revealed that `GetDrive` was failing to parse one of our drives that lists its `Identifiers` as an object instead of an array. Most of the drives do not do this but the returned drives slice was quite small because `chassis.Drives()` [returns early](https://github.com/stmcginnis/gofish/blob/main/redfish/chassis.go#L375) when it encounters an error for any drive in the list.

```yaml
  "Identifiers": {
    "DurableName": "5CD2E4D40DC20100",
    "DurableNameFormat": "EUI"
  },
```

Is this something you have encountered before? I don't know how to parse the variable type besides using an interface and type assertions but that seems hacky.

stmcginnis:
I have not run in to that, but it definitely is not following the spec. It explicitly states `"type": "array"`, so this looks like another case (unfortunately) that will have to go back to the vendor to get them to fix their implementation.

We could certainly change those instances to just capture the error but continue on through the list. I would have thought in most cases things would either fail right away or be fine, but this appears to be at least one case where maybe there are some legitimately formatted results.

If you step through debugging, does it actually return some valid drive objects before hitting the malformed one? If so, feel free to propose a change to not return early from that loop. I think ideally then we should do the same for all instances of that pattern where we are gathering up a collection of objects, so take it however far you want to. If you don't have the time/inclination to change them all, then I can follow up to get the rest.

Or if not at all, I can put it on my list to update sometime in the near future. :)

Thanks for all the detail @sosodev!

sosodev:
Thanks for pointing out that this is vendor jank. Yeah, it was hitting some valid drive entities before the malformed one. I'll start working on a PR to collect errors rather than return early. I suppose you we don't want to change the signature of those funcs, right? Any idea what should be done with the collected errors?

stmcginnis:
Good question. Maybe something like `fmt.Errorf("errors retrieving items: %v", collectedErrors)`? (not saying exactly that, but hopefully you get the idea)

sosodev:
Hmm, I was just thinking about this today when I realized a slight problem. Collecting and returning the errors merged into a single error does subtly change the function signature. Previously any errors in the loop were ignored and only critical errors were returned. Existing code that uses these functions would need to be modified to handle the new return case of collection + soft error.

stmcginnis:
[...]

PR #150, opened 2021-05-26
Collect errors in a custom error type when fetching a collection

sosodev:
Relevant discussion in #147 

I updated all of the collection functions that I could find by searching for `var result []`. I appreciate the consistency of the library because I think that returned most of the collection functions. :smile: Let me know if I missed any.

I noticed while working through these changes that the case of `return result, nil` happened much less often than I thought so my worry about user code brittleness is less valid. :+1: 

Overall I think this change improves the consistency of collection error handling and should greatly improve the resilience of gofish in production. Let me know what you think!

Edit: 
Also, I chose to implement the bulk of the error message with JSON because I figure that some of the error messages could end up being quite large. I think it will be nice to format the error message and have a clear mapping of errors to the entities with which they occurred.

### DIFF
--- a/common/collection.go
+++ b/common/collection.go
@@ -58,19 +58,20 @@ func GetCollection(c Client, uri string) (*Collection, error) {
 	return &result, nil
 }
 
-type collectionError struct {
+// CollectionError is used for collecting errors when working with collections
+type CollectionError struct {
 	Failures map[string]error
 }
 
-// NewCollectionError gets you a new *collectionError
+// NewCollectionError gets you a new *CollectionError
 // it's useful for collecting and formatting errors that occur when fetching a collection
-func NewCollectionError() *collectionError {
-	return &collectionError{
+func NewCollectionError() *CollectionError {
+	return &CollectionError{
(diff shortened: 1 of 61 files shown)


sosodev:
Testing this update with some of our tooling and it's a dramatic improvement on a machine with a bad drive early in the collection. From around 5 drives returned to all but the bad drive returned with an error message that specifically points out the bad drive. :)

stmcginnis:
Thanks @sosodev, overall this looks great. Can you address the lint job failures? Might just need a `// nolint` tag on some of them. You can run `make lint` locally to test.

sosodev:
Hi, I'll get this update soon. Sorry about the delay. It's hard to keep up sometimes. :slightly_smiling_face:

stmcginnis:
No worries at all @sosodev! Thanks for working on it. :)

sosodev:
Hi @stmcginnis pretty sure I've fixed all of the lint errors now.

stmcginnis (approved):
Awesome, thanks!

PR #151, opened 2021-06-07
Manager set to plugin system

mikeletux:
Hi Sean,

I need to use oem, oemLinks and OemActions on my Terraform provider. I did the same as the simpleUpdate for the plugin system.

Changed oem and oemLinks from map[string]interface{} to json.RawMessage, so everyone can build structs around it and parse that stuff accordingly.

Also I added an OemActions field, so OemActions can be parsed later as well.

Let me know!
Best regards.
/Miguel


stmcginnis (approved):
Looks good, thanks!

PR #153, opened 2021-06-18
add oem property for root service object

leslie-qiwa:

### DIFF
--- a/serviceroot.go
+++ b/serviceroot.go
@@ -99,6 +99,9 @@ type Service struct {
 	// Managers shall only contain a reference to a collection of resources that
 	// comply to the Managers schema.
 	managers string
+	// Oem contains all the vendor specific actions. It is vendor responsibility to parse
+	// this field accordingly
+	Oem json.RawMessage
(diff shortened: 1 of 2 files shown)


stmcginnis:
Thanks @leslie-qiwa, but please see https://github.com/stmcginnis/gofish/pull/146 for an example of how we've decided to approach OEM data.

leslie-qiwa:
@stmcginnis let me know if this new change is align with #146 or not

serviceroot.go
```diff
@@ -175,6 +178,7 @@ func (serviceroot *Service) UnmarshalJSON(b []byte) error {
 		Links              struct {
 			Sessions common.Link
 		}
+		Oem json.RawMessage // OEM message will be stored here
```
stmcginnis, line 181: I don't think we would need this because it can just be unmarshalled directly to the object's property. But no big deal, that can be cleaned up later.

stmcginnis (approved):
Looks good, thanks!

PR #165, opened 2022-02-25
fix type of prefix length of ipv6 address

leslie-qiwa:
range of prefix length of ipv6 address should be 1-128.


stmcginnis (approved):
Looks good - thanks!

PR #166, opened 2022-03-11
redfish/updateservice: struct field MultipartHTTPPushURI

joelrebel:
This field enables clients to determine if the updateservice supports Multipart push updates.

_A service may support the `MultipartHttpPushUri` property within the
UpdateService resource. A client can perform an HTTP or HTTPS POST request on the URI specified by this property to initiate a push-based update._

https://www.dmtf.org/sites/default/files/standards/documents/DSP2062_1.0.0.pdf


stmcginnis (approved):
Thanks!

PR #167, opened 2022-03-11
client: export method runRawRequestWithHeaders

joelrebel:
To enable clients to submit their own multipart payload,
it would help to export the `runRawRequestWithHeaders` method.

In my case, the multipart form-data needed to be in the following format,
which isn't directly supported within the `runRequestWithMultipartPayloadWithHeaders` 
method, and needed more work to prepare the first part of the MIME header.

```
Content-Length: 416
Content-Type: multipart/form-data; boundary=--------------------
----1771f60800cb2801

--------------------------1771f60800cb2801
Content-Disposition: form-data; name="UpdateParameters"
Content-Type: application/json

{"Targets": [], "@Redfish.OperationApplyTime": "OnReset", "Oem":
{}}
--------------------------1771f60800cb2801
Content-Disposition: form-data; name="UpdateFile"; filename="dum
myfile"
Content-Type: application/octet-stream

hey.
--------------------------1771f60800cb2801--
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -381,6 +381,11 @@ func (c *APIClient) runRawRequest(method, url string, payloadBuffer io.ReadSeeke
 	return c.runRawRequestWithHeaders(method, url, payloadBuffer, contentType, nil)
 }
 
+// RunRawRequestWithHeaders actually performs the REST calls but allowing custom headers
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Makes sense, thanks!

PR #168, opened 2022-04-07
fix(simple-storage/device): add device name

mahuihuang:
Add a name field to distinguish between devices.

### DIFF
--- a/redfish/simplestorage.go
+++ b/redfish/simplestorage.go
@@ -12,6 +12,8 @@ import (
 
 // Device shall describe a storage device visible to SimpleStorage.
 type Device struct {
+	// Name shall be a user-friendly name for the device.
+	Name string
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks, looks good.

stmcginnis:
Looks like I need to update the linter rules. Failure is not related to this change.

mahuihuang:
Excuse me, when will the next stable version be released? I have a project that uses this dependency.

stmcginnis:
You can always reference by sha in your `go.mod`, but I can tag an actual release if that makes it easier.

stmcginnis:
Published: https://github.com/stmcginnis/gofish/releases/tag/v0.13.0

mahuihuang:
Thank you for your reply

PR #169, opened 2022-04-25
Add GetService() method to APIClient

micahhausler:
This enables callers to write testable interfaces instead of directly accessing fields on the APIClient.

For example
```go

// defined in a caller's package
type RedfishClient interface {
    GetService() *Service
    Logout()
}

type MyType struct {
    client RedfishClient 
}

func NewMyType(endpoint string) (*MyType, error) {
    r := &MyType{}
    var err error
    r.client, err = gofish.ConnectDefault(endpoint)
    if err != nil {
        return nil, err
    }
    return r, nil
)

func main(){
    mt, _ := NewMyType("")
    fmt.Println(mt.client.GetService().Systems()[0].BIOSVersion)
    mt.client.Logout()
}

```
Signed-off-by: Micah Hausler <mhausler@amazon.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -218,6 +218,11 @@ func ConnectDefaultContext(ctx context.Context, endpoint string) (c *APIClient, 
 	return client, err
 }
 
+// GetService returns the APIClient's service.
+func (c *APIClient) GetService() *Service {
+	return c.Service
+}
+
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Lint job failure is unrelated. This looks good and makes sense - thanks!

PR #182, opened 2022-06-07
client: set custom header Content-Length

joelrebel:
The Go http client Do() method ignores the Content-Length header when set with `Header.Set()`.
This change enables the caller to specify a custom Content-Length for cases where the caller is sure about the payload size - multipart payloads where the file contents are not buffered.


stmcginnis (approved):
Thanks @joelrebel!

Issue #183, opened 2022-06-10
feat: support bootoption resources


bzub:
`BootOption`s are needed for my use-case of mapping boot device paths to boot references, allowing me to modify the boot order more easily across platforms.

PR #184, opened 2022-06-10
feat: add bootoption resource

bzub:
Fixes: https://github.com/stmcginnis/gofish/issues/183

This is my first look at this codebase so please let me know if anything is needed.
Thanks!

### DIFF
--- a/redfish/bootoption.go
+++ b/redfish/bootoption.go
@@ -1,0 +1,86 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package redfish
+
+import (
+	"encoding/json"
+
+	"github.com/stmcginnis/gofish/common"
+)
+
+// BootOption represents the properties of a bootable device available in the
+// system.
+type BootOption struct {
+	common.Entity
+
+	// ODataContext is the odata context.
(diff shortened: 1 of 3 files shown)


redfish/bootoption.go
```diff
@@ -0,0 +1,86 @@
...
+	// BootOptionReference is the unique identifier seen in Boot.BootOrder.
+	BootOptionReference string
+	// DisplayName is the user-readable display name of the boot option
+	// that appears in the boot order list in the user interface.
+	DisplayName string
+	// UefiDevicePath is the UEFI device path to access this UEFI boot
+	// option.
+	UefiDevicePath string
+	// rawData holds the original serialized JSON so we can compare updates.
+	rawData []byte
```
stmcginnis, line 39: Looks like `rawData`, `etag`, and specialized `UnmarshalJSON` could all be removed since this is a simple object that does not have any of its own actions or specially serialized fields.

We could probably clean that up later though, and really no harm having it present other than a very slight extra overhead.

stmcginnis (approved):
Looks good, thanks for adding this.

stmcginnis (changes requested):
Looks like there are some linting errors to be addressed first.

PR #178, opened 2022-06-02
feat: support oem field of thermal and fan

daadajuan:

PR #186, opened 2022-06-30
Support event subscription and submitTestEvent for various vendors

nwaizer:
Hi Sean,
I seek to add support for Dell, HPE and ZT systems in their event subscription and SubmitTestEvent.
My code is tested on all these vendors.
Any feedback is welcome.
NIv

### DIFF
--- a/oem/dell/dell.go
+++ b/oem/dell/dell.go
@@ -1,0 +1,56 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package dell
+
+import (
+	"fmt"
+	"github.com/stmcginnis/gofish/common"
+	"net/http"
+)
+
+const (
+	eventContext string = "root"
+)
+
+var (
+	submitTestEventTarget = "/redfish/v1/EventService/Actions/EventService.SendTestEvent"
+)
+
+type (
+	dellPayloadType struct {
+		Destination string `json:"Destination"`
+		EventTypes  string `json:"EventTypes"`
+		Context     string `json:"Context"`
+		Protocol    string `json:"Protocol"`
+		MessageID   string `json:"MessageId"`
+	}
+)
+
+// SendEventDell sends event according to msgId and returns error.
+func SendEventDell(client common.Client, messageID, eType, protocol string) error {
+	payload := dellPayloadType{
+		Destination: submitTestEventTarget,
+		EventTypes:  eType,
+		Context:     eventContext,
+		Protocol:    protocol,
+		MessageID:   messageID,
+	}
+	resp, err := client.Post(submitTestEventTarget, payload)
+
+	if err != nil {
+		return fmt.Errorf("failed to post submitTestEvent due to: %w", err)
+	}
+	defer resp.Body.Close()
+
+	valid := map[int]bool{
+		http.StatusNoContent: true,
+		http.StatusCreated:   true}
+
+	if !valid[resp.StatusCode] {
+		return fmt.Errorf("on send event received response: %v due to: %s", resp.StatusCode, resp.Body)
+	}
+
+	return nil
+}
--- a/oem/zt/zt.go
+++ b/oem/zt/zt.go
@@ -1,0 +1,98 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package zt
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+import (
+	"github.com/stmcginnis/gofish/common"
+)
+
+var (
+	SubmitTestEventTarget = "/redfish/v1/EventService/Actions/EventService.SendTestEvent"
+)
+
+type (
+	SubscriptionZtRequestType struct {
+		Destination string            `json:"Destination"`
+		HTTPHeaders map[string]string `json:"HttpHeaders,omitempty"`
+		Protocol    string            `json:"Protocol,omitempty"`
+		Context     string            `json:"Context,omitempty"`
+	}
+
+	// subscribeResponseType zt uses a unique subscription response.
+	subscribeZtResponseType struct {
+		ODataContext         string `json:"@odata.context"`
+		ODataEtag            string `json:"@odata.etag"`
+		ODataID              string `json:"@odata.id"`
+		ODataType            string `json:"@odata.type"`
+		Context              string `json:"Context"`
+		DeliveryRetryPolicy  string `json:"DeliveryRetryPolicy"`
+		Description          string `json:"Description"`
+		Destination          string `json:"Destination"`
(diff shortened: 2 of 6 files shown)


stmcginnis (changes requested):
Thanks for proposing this @nwaizer! OEM handling is handled by converting the standard objects over to their OEM equivalents to keep OEM specific things out of the common code. Please take a look at https://github.com/stmcginnis/gofish/pull/178 for an example of how this is done.

If we can get this moved into ZT-specific OEM code, this would be great to add.

nwaizer:
Thank you @stmcginnis is this commit what you meant?

nwaizer:
@stmcginnis I got a few question about #178 : 
1. According to the pull request, I am committing three files under the oem directory, for the three vendors I tested. 
2. For each vendor I need to implement these two redfish API calls:
> Subscribe to events
> SubmitTestEvent - to simulate an event
3. As each vendor implements its own payload for these requests and its own response, I also added the necessary structures for them in the files.
4. Your implementation of SubmitTestEvent() need to switch the call according to the oem vendor, so in case it is Dell is will call SendEvnetDell()

Whats next?
Niv

redfish/eventdestination.go
```diff
@@ -7,6 +7,7 @@ package redfish
 import (
 	"encoding/json"
 	"fmt"
+	oemZt "github.com/stmcginnis/gofish/oem/zt"
```
stmcginnis, line 10: Common code should not import the OEM code.

oem/dell/dell.go
```diff
@@ -0,0 +1,52 @@
+package dell
```
stmcginnis, line 1: Can you add the license header to these files?

oem/dell/dell.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"github.com/stmcginnis/gofish/redfish"
```
stmcginnis, line 5: Please move this down into its own local import group.

oem/zt/zt.go
```diff
@@ -0,0 +1,96 @@
...
+	subscriptionPayloadZT struct {
+		Destination string                           `json:"Destination"`
+		HTTPHeaders map[string]string                `json:"HttpHeaders,omitempty"`
+		Protocol    redfish.EventDestinationProtocol `json:"Protocol,omitempty"`
+		Context     string                           `json:"Context,omitempty"`
+	}
+
+	// subscribeResponseType zt uses a unique subscription response.
+	subscribeResponseType struct {
+		OdataContext        string `json:"@odata.context"`
```
stmcginnis, line 26: To keep this consistent with the rest of the library, can you make these "OData" instead of "Odata"?

https://github.com/stmcginnis/gofish/blob/main/common/types.go#L37

oem/zt/zt.go
```diff
@@ -0,0 +1,96 @@
...
+		Destination         string `json:"Destination"`
+		EventFormatType     string `json:"EventFormatType"`
+		ID                  int    `json:"ID"`
+		Name                string `json:"Name"`
+		Protocol            string `json:"Protocol"`
+		Status              struct {
+			Health       string `json:"Health"`
+			HealthRollup string `json:"HealthRollup"`
+			State        string `json:"State"`
+		}
```
stmcginnis, line 42: This could just use the `Status` struct: https://github.com/stmcginnis/gofish/blob/main/common/types.go#L261

redfish/eventservice.go
```diff
@@ -7,11 +7,13 @@ package redfish
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stmcginnis/gofish/common"
+	oemDell "github.com/stmcginnis/gofish/oem/dell"
+	oemHp "github.com/stmcginnis/gofish/oem/hpe"
+	oemZt "github.com/stmcginnis/gofish/oem/zt"
```
stmcginnis, line 13: Please remove circular relationships.

stmcginnis (commented):
Thanks. Still some work needed to separate the OEM functionality from the core standard parts. Also looks like there is now a merge conflict that will need to be resolved.

redfish/eventdestination.go
nwaizer: The function [CreateEventSubscription](https://github.com/stmcginnis/gofish/blob/main/redfish/eventservice.go#L302) is the common call to be used on all three vendors.
This function warps the [CreateEventDestination](https://github.com/stmcginnis/gofish/blob/46c07e7c05bda4b260b98647ab3c379c627bdd87/redfish/eventdestination.go#L342) which uses a http request payload that ZT systems does not accept.
ZT systems only accepts this [payload](https://github.com/nwaizer/gofish/blob/ZTsystem/oem/zt/zt.go#L17)

The [oem check](https://github.com/stmcginnis/gofish/blob/46c07e7c05bda4b260b98647ab3c379c627bdd87/redfish/eventdestination.go#L377) only adds the oem data to the existing payload.
I need a way to allow me to change the payload in case it is the vendor ZT, before you issue a post with it.

So I used:
```
	// ZT systems uses a different payload for this action
	if oem == "ZT" {
		return oemZt.SubscribeZT(c, uri, destination, protocol, context)
	}
```

oem/dell/dell.go
nwaizer: done

oem/dell/dell.go
nwaizer: done

oem/zt/zt.go
nwaizer: done

oem/zt/zt.go
nwaizer: done

redfish/eventservice.go
nwaizer: SubmitTestEvent() is implemented differently for three oem vendors. They use different http request payloads and respond with different "OK" status codes.

The current (SubmitTestEvent)[https://github.com/stmcginnis/gofish/blob/main/redfish/eventservice.go#L334) function does not work for all vendors. I would ask to use the [SendTestEvent](https://github.com/nwaizer/gofish/blob/ZTsystem/redfish/eventservice.go#L377) in order to send this request with the payload each vendor accepts.

stmcginnis: Same as above, implementations for the OEM specific calls need to move into the OEM specific code.

nwaizer: circular relationships removed in updated commit

redfish/eventdestination.go
nwaizer: will do after answer for https://github.com/stmcginnis/gofish/pull/186#discussion_r913713326

nwaizer:
need to create new PR due to commit error

PR #189, opened 2022-07-10
hpe events test

nwaizer:
I added events for hpe as well as a test, will real captured eventServiceBody.
But now when I invoke the tests I get this error:
$ go test -v oem/hpe/eventservice_test.go 
# command-line-arguments [command-line-arguments.test]
oem/hpe/eventservice_test.go:217:13: undefined: EventService
oem/hpe/eventservice_test.go:289:45: undefined: PayloadType
oem/hpe/eventservice_test.go:318:26: undefined: FromEventService
FAIL	command-line-arguments [build failed]
FAIL

$ go test -v oem/dell/eventservice_test.go 
# command-line-arguments [command-line-arguments.test]
oem/dell/eventservice_test.go:156:13: undefined: EventService
oem/dell/eventservice_test.go:228:45: undefined: PayloadType
oem/dell/eventservice_test.go:258:27: undefined: FromEventService
FAIL	command-line-arguments [build failed]
FAIL

Yet all these worked fine yesterday, and the editor sees no mistake, what am I missing?

### DIFF
--- a/oem/hpe/eventservice.go
+++ b/oem/hpe/eventservice.go
@@ -1,0 +1,79 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package hpe
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/stmcginnis/gofish/redfish"
+)
+
+type PayloadType struct {
+	EventID           string `json:"EventID"`
+	EventTimestamp    string `json:"EventTimestamp"`
+	EventType         string `json:"EventType"`
+	Message           string
+	MessageArgs       []string
+	MessageID         string `json:"MessageId"`
+	OriginOfCondition string
+	Severity          string
+}
+
+// EventService is the Hpe-specific handler for the EventService instance.
+type EventService struct {
+	redfish.EventService
+}
+
+// SubmitTestEvent sends event according to msgId and returns error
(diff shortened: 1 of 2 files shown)


stmcginnis:
There were some other updates that were not done to the HPE one since it was removed in the cleanup. Why does it need to be readded? When looking at the previous code, it appears from the code and their docs that HPE is actually following the standard, so there wasn't a need to have an OEM-specific implementation added.

nwaizer:
1. [SubmitTestEvent](https://github.com/stmcginnis/gofish/blob/main/redfish/eventservice.go#L335) is diffrent from my offered function:
- response status code is checked, so even if POST doe snot give an error, the response status code, maybe wrong. several responses are good.
- the payload field called Severity differs
- I want to sent the messageID, as in other vendors, its easier to compare, as all use it.
- EventTimestamp shoudl be RFC3339 format fro my testing.
2. test uses mock and actual captured responses

stmcginnis:
It sounds like all of that should be addressed in the core `SubmitTestEvent` implementation.

It would be good to add a Status Code check.

Severity is the same - it is a string according to the Redfish specification and that is what you have it defined as in this proposed change.

That would make sense to change the method signature to add the ability to pass in things like the `MessageId`.

The spec isn't entirely clear on expected timestamp format, but looking at what is in the spec there is an example request given that does show `EventTimestamp` formatted with an RFC3339 timestamp.

PR #192, opened 2022-08-03
Add workaround to ZT systems rate limit bug.

nwaizer:


stmcginnis:
Thanks, but pulling in all of apimachinery is pretty heavy wait for this. Is there another library that can give the functionality you need without doing that? Gofish dependencies should be kept very minimal.

nwaizer:
may I use github.com/avast/retry-go ?

stmcginnis:
The retry needs are pretty basic, so it really would be best to just put in a retry loop and not bring in a whole library just to do it. Something along the lines of:

```go
for i := 0; i < retryAttempts; i++ {
    resp, err = eventservice.Client.Post(eventservice.SubmitTestEventTarget, p)
    if err == nil {
        break
    }

    log.Printf("Failed to send test event (%d/%d)\n", i+1, retryAttempts)
    time.Sleep(retryInterval)
}

if err != nil {
    return err
}
```

nwaizer:
Done

go.mod
```diff
@@ -1,3 +1,9 @@
 module github.com/stmcginnis/gofish
 
 go 1.16
+
+require (
```
stmcginnis, line 5: These changes to `go.mod` and `go.sum` still need to be reverted.

oem/zt/eventservice.go
```diff
@@ -80,15 +82,47 @@ func (eventservice *EventService) Subscribe(eventsReceiverURL string, protocol r
...
+		if err == nil {
+			if retryCounter > retryReportThreshold {
+				log.Printf("Had to retry %v times to send SubmitTestEvent()", retryCounter)
+			}
+
+			break
+		}
+
+		if resp != nil {
+			err = resp.Body.Close()
```
stmcginnis, line 115: Is there really a need to fail the operation if something like this happens? I don't think it possibly could, but I think instead of returning an error it would be fine to just do:

```go
if resp != nil {
    _ = resp.Body.Close()
}
time.Sleep(retryInterval)
```

stmcginnis (changes requested):

go.mod
nwaizer: done

stmcginnis (approved):

PR #193, opened 2022-08-08
Add oem query

nwaizer:


go.mod
```diff
@@ -1,3 +1,9 @@
 module github.com/stmcginnis/gofish
 
 go 1.16
+
+require (
```
stmcginnis, line 5: Pulling in dependencies should be avoided. We want to keep this library as light weight as possible, so effort has been made to avoid adding anything.

oem/oem.go
```diff
@@ -0,0 +1,62 @@
...
+
+	switch {
+	case strings.Contains(oem, DellRedfishOem):
+		return Dell, nil
+	case strings.Contains(oem, HpeRedfishOem):
+		return Hpe, nil
+	case strings.Contains(oem, ZTRedfishOem):
+		return ZT, nil
+	default:
+		return "", fmt.Errorf("failed to match vendor from output: %v", oem)
```
stmcginnis, line 46: It's possible for someone to "bring their own" OEM library (which was actually my original intention).

This is a decent idea to have a factory for it, but the user would still need to look up which vendor is being used to be able to tell what library and objects they have. The return on line 51 below is a problem. It works now because you are only returning Dell, but there doesn't appear to be a good way to have this return whatever vendor specific OEM implementation it is for any of the included OEMs.

It's an interesting idea, but I'd rather keep the library simple and put the OEM special handling on the consumer since there are a lot of decisions and other concerns they will need to make and be aware of when accessing the OEM-specific things.

stmcginnis (changes requested):
Some thoughts commented in `oem/oem.go`. I don't think we want to merge this, but I am going to keep it open for a bit to see if anyone else wants to join the discussion and has ideas how to make this a better abstraction.

Thanks for working on it and looking for ways to make things easier to use, but unfortunately I don't think this is something we can do cleanly.

stmcginnis:
Closing based on the above concerns. Thanks for looking in to this, but I think we'll need a slightly different approach.

PR #194, opened 2022-08-15
Bugfix/remove http header cookie

Sn0rt:
1. remove cookie ( the header can set by manually
2. bugfix: only set etag which query from /bios/setting


stmcginnis (approved):
Looks good. Wasn't sure about dropping the cookie, but noticed this explicit statement in the redfish spec:

> Shall not use HTTP cookies to authenticate any activity, such as GET , POST , PUT , PATCH , and DELETE .

Sn0rt:
> Looks good. Wasn't sure about dropping the cookie, but noticed this explicit statement in the redfish spec:
> 
> > Shall not use HTTP cookies to authenticate any activity, such as GET , POST , PUT , PATCH , and DELETE .

the user can add this header manually

PR #198, opened 2022-09-30
Fixed issues and updated Redfish schema version in tools/

pallasscat:
Fixed newlines in tools/source.tmpl; fixed issues and updated Redfish schema version in tools/gen_script.sh


pallasscat:
I guess that's all issues fixed now >>

PR #199, opened 2022-10-03
Various fixes in tools/ to improve code generation

pallasscat:
Fixed formatting issues (double spaces, excessive newlines, const names) in template output; fixed directory issues and cleaned up code in gen_script.sh; switched to int-based schema version selection and `re` for code cleanup, minor fixes in generate_from_schema.py.


pallasscat:
Additionally: do we still need the full template? For example, UnmarshalJSON() method is unnecessary more often than not.

Alternatively, how about re-writing the entire generate.py + schema.tmpl to render `enum`s and `object`s exclusively?

PR #204, opened 2022-10-20
Update ListReferencedMemorys to recover the memories asynchronously

lorenzb07:
For my use case, I need to recover many memories info and the method ListReferencedMemorys is too slow beacuse calls the GetMemory sequentially. 
This pull request recovers the memories using channel and goroutine.

### DIFF
--- a/redfish/memory.go
+++ b/redfish/memory.go
@@ -7,6 +7,7 @@ package redfish
 import (
 	"encoding/json"
 	"reflect"
+	"sync"
 
 	"github.com/stmcginnis/gofish/common"
 )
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Awesome!

stmcginnis:
Issue with the linting job, but I can clean that up separately.

lorenzb07:
In this pull request there is an error. 
A `wg.Add(1)` is missing.

The new pull request that fix this is [here (208)](https://github.com/stmcginnis/gofish/pull/208)

PR #209, opened 2022-10-21
Fix get memeories async

lorenzb07:
this pull request fix the [204](https://github.com/stmcginnis/gofish/pull/204) adding the wg.Add(1) in the for loop and fix the collectionError


stmcginnis (approved):
Good catch, I missed that in review too.

PR #210, opened 2022-10-24
Make collections retrieval parallel

stmcginnis:
This makes calls to get the objects from a collection link parallel.

Some collections can contain a large number of items to retrieve. Doing this serially can have a noticable performance impact. This change allows three fetches in parallel so we don't overwhelm the Redfish service and cause a denial of service. With three concurrent operations we can be processing some responses while queuing up the more, making it a lot faster to process.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,10 +31,6 @@ linters-settings:
     min-complexity: 25
   goimports:
     local-prefixes: github.com/stmcginnis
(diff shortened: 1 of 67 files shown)


stmcginnis:
Been out here long enough. Was hoping for external validation that things are working well, but I'm sure I'll find out soon enough if there are issues. ;)

PR #213, opened 2022-10-26
feat(oem/supermicro): Add Supermicro Manager OEM functions

feuerrot:
This PR includes functions for:
* IKVM (Java/HTML5)
* NTP (Enabled/Servers/DST)
* Syslog (Enabled/Server)

I'd appreciate any hints regarding the `supermicro.Manager`-structure and how custom manager extensions should be integrated - while it does work accessing the syslog configuration using `smm.supermicroManager.Oem.Supermicro.Syslogs()` seems to be rather awkward.

### DIFF
--- a/oem/supermicro/ikvm.go
+++ b/oem/supermicro/ikvm.go
@@ -1,0 +1,78 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package supermicro
+
+import (
+	"encoding/json"
+	"reflect"
+
+	"github.com/stmcginnis/gofish/common"
+)
+
+type IKVMInterface string
+
+const (
+	JavaIKVMInterface IKVMInterface = "JAVA plug-in"
+	HTMLIKVMInterface IKVMInterface = "HTML 5"
+)
+
+type IKVM struct {
+	common.Entity
+
+	OdataType        string        `json:"@odata.type"`
+	CurrentInterface IKVMInterface `json:"Current interface"`
+	URI              string        `json:"URI"`
+
+	rawData []byte
+}
+
+func (ikvm *IKVM) UnmarshalJSON(b []byte) error {
+	type temp IKVM
+
+	var rtn temp
+	err := json.Unmarshal(b, &rtn)
+	if err != nil {
+		return err
+	}
+
+	*ikvm = IKVM(rtn)
+	ikvm.rawData = b
+
+	return nil
+}
+
+func (ikvm *IKVM) Update() error {
+	original := new(IKVM)
+	err := original.UnmarshalJSON(ikvm.rawData)
+	if err != nil {
+		return err
+	}
+
+	readWriteFields := []string{
+		"Current interface",
+	}
+
+	originalElement := reflect.ValueOf(original).Elem()
+	currentElement := reflect.ValueOf(ikvm).Elem()
+
+	return ikvm.Entity.Update(originalElement, currentElement, readWriteFields)
+}
+
+func GetIKVM(c common.Client, uri string) (*IKVM, error) {
+	resp, err := c.Get(uri)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var ikvm IKVM
+	err = json.NewDecoder(resp.Body).Decode(&ikvm)
+	if err != nil {
+		return nil, err
+	}
+
+	ikvm.SetClient(c)
+	return &ikvm, nil
+}
--- a/oem/supermicro/ikvm.go
+++ b/oem/supermicro/ikvm.go
@@ -21,7 +21,7 @@ const (
 type IKVM struct {
 	common.Entity
 
-	OdataType        string        `json:"@odata.type"`
+	ODataType        string        `json:"@odata.type"`
 	CurrentInterface IKVMInterface `json:"Current interface"`
 	URI              string        `json:"URI"`
 
--- a/oem/supermicro/manager.go
+++ b/oem/supermicro/manager.go
@@ -40,13 +40,13 @@ func FromManager(manager *redfish.Manager) (Manager, error) {
 }
 
 func (manager *Manager) NTP() (*NTP, error) {
-	return GetNTP(manager.Client, string(manager.Oem.Supermicro.NTP))
+	return GetNTP(manager.GetClient(), string(manager.Oem.Supermicro.NTP))
 }
 
 func (manager *Manager) Syslog() (*Syslog, error) {
-	return GetSyslog(manager.Client, string(manager.Oem.Supermicro.Syslog))
+	return GetSyslog(manager.GetClient(), string(manager.Oem.Supermicro.Syslog))
 }
 
 func (manager *Manager) IKVM() (*IKVM, error) {
-	return GetIKVM(manager.Client, string(manager.Oem.Supermicro.IKVM))
+	return GetIKVM(manager.GetClient(), string(manager.Oem.Supermicro.IKVM))
 }
--- a/oem/supermicro/manager.go
+++ b/oem/supermicro/manager.go
@@ -1,0 +1,53 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
+package supermicro
+
+import (
+	"encoding/json"
+
+	"github.com/stmcginnis/gofish/common"
+	"github.com/stmcginnis/gofish/redfish"
+)
+
+type Manager struct {
+	redfish.Manager
+	Oem ManagerOem
+}
+
+type ManagerOem struct {
+	Supermicro ManagerOemSupermicro
+}
+
+type ManagerOemSupermicro struct {
+	OdataType string `json:"@odata.type"`
+	NTP       common.Link
+	Syslog    common.Link
+	IKVM      common.Link
+	m         *redfish.Manager
+}
+
+func FromManager(manager *redfish.Manager) (Manager, error) {
+	var oem ManagerOem
+	_ = json.Unmarshal(manager.Oem, &oem)
+
+	oem.Supermicro.m = manager
+
+	return Manager{
+		Manager: *manager,
+		Oem:     oem,
+	}, nil
+}
+
+func (manager *ManagerOemSupermicro) NTPs() (*NTP, error) {
+	return GetNTP(manager.m.Client, string(manager.NTP))
+}
+
+func (manager *ManagerOemSupermicro) Syslogs() (*Syslog, error) {
+	return GetSyslog(manager.m.Client, string(manager.Syslog))
+}
+
+func (manager *ManagerOemSupermicro) IKVMs() (*IKVM, error) {
+	return GetIKVM(manager.m.Client, string(manager.IKVM))
+}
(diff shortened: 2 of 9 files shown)


stmcginnis:
Awesome!

> while it does work accessing the syslog configuration using smm.supermicroManager.Oem.Supermicro.Syslogs() seems to be rather awkward.

The way it was invisioned for these OEM extensions it would be a matter of creating the OEM type from the standard type. So something roughly along these lines:

```go
manager := c.GetManager()

smManager := supermicro.FromManager(manager)
for _, logs := range smManager.SysLogs() {
    // process
}
```

So rather than trying to mirror the whole structure, you just hydrate a new OEM object using the existing Redfish object. A decent example is probably here: https://github.com/stmcginnis/gofish/blob/main/oem/dell/eventservice.go

feuerrot:
@stmcginnis If I understand the Redfish Specification correctly, Supermicros Manager OEM extensions are separate resouces and not part of the manager itself, while Dells Eventservice contains the extensions directly. Does this make a difference in how the Supermicro extensions should be accessed from the manager?

stmcginnis:
> @stmcginnis If I understand the Redfish Specification correctly, Supermicros Manager OEM extensions are separate resouces and not part of the manager itself, while Dells Eventservice contains the extensions directly. Does this make a difference in how the Supermicro extensions should be accessed from the manager?

Thanks for the details. That may change a little of how the resulting structs are defined, but that wouldn't change how they are accessed. All OEM data should be accessed by using the standard Redfish objects, then using an OEM implementation to access the `oem` parts to handle vendor specific (i.e. non-standard) paths.

feuerrot:
I modified the manager accordingly.

oem/supermicro/ikvm.go
```diff
@@ -0,0 +1,78 @@
...
+	common.Entity
+
+	OdataType        string        `json:"@odata.type"`
+	CurrentInterface IKVMInterface `json:"Current interface"`
+	URI              string        `json:"URI"`
+
+	rawData []byte
+}
+
+func (ikvm *IKVM) UnmarshalJSON(b []byte) error {
```
stmcginnis, line 31: It doesn't look like this or `Update` are needed since there is nothing special needed for deserialization for JSON and there are no updatable fields.

No harm having them though.

oem/supermicro/manager.go
```diff
@@ -0,0 +1,52 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/stmcginnis/gofish/common"
+	"github.com/stmcginnis/gofish/redfish"
+)
+
+type Manager struct {
+	redfish.Manager
+	Oem ManagerOem
```
stmcginnis, line 17: It might be a simple interface to just get rid of this `Oem` field and move the fields of `ManagerOem` up a level into `Manager`.

oem/supermicro/manager.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type ManagerOem struct {
+	Supermicro struct {
+		OdataType string `json:"@odata.type"`
+		NTP       common.Link
```
stmcginnis, line 23: Then make these fields lower case. The links aren't needed by the user of the library, just the methods being added below to access the objects that those links point to.

So combined with the above, it would just be something like:

```go
smManager, err := supermicro.FromManager(redfishManager)
if err != nil {
    // Handle it
}

ntp, err := smManager.NTP()
```

stmcginnis (commented):
Looks great! I've added some suggestions, so I'll hold off on merging until you've had a chance to take a look and respond.

oem/supermicro/ikvm.go
feuerrot: The `CurrentInterface` is changeable and we use it to change the setting from Java to HTML. This populates the URI which can be used without further authentication: 

```go
func (barsch *Barsch) GetIKVM() error {
	ikvm, err := barsch.Manager.IKVM()
	if err != nil {
		return fmt.Errorf("IKVM: %v", err)
	}

	if ikvm.CurrentInterface != supermicro.HTMLIKVMInterface {
		ikvm.CurrentInterface = supermicro.HTMLIKVMInterface
		if err := ikvm.Update(); err != nil {
			return fmt.Errorf("can't switch to HTML5 IKVM: %v", err)
		}
		ikvm, err = barsch.Manager.IKVM()
		if err != nil {
			return fmt.Errorf("can't update IKVM after switch to HTML Interface: %v", err)
		}
	}

	log.Printf("URI=%s%s", barsch.hostname, ikvm.URI)

	return nil
}
```

(Current versions of the X12 IPMI firmware will return a 404, but it did work on older ones - I'll probably open a bug report with supermicro about it)

oem/supermicro/ikvm.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	JavaIKVMInterface IKVMInterface = "JAVA plug-in"
+	HTMLIKVMInterface IKVMInterface = "HTML 5"
+)
+
+type IKVM struct {
+	common.Entity
+
+	OdataType        string        `json:"@odata.type"`
```
stmcginnis, line 24: We've standardized on `OData` (versus `Odata`) elsewhere. Mind changing these to be consistent?

stmcginnis (changes requested):
Sorry for taking ridiculously long to get back to this. There are a couple comments still not addressed. If you can make a few updates, this looks really good and we can get it merged.

Thanks for working on this!

feuerrot:
Hi,
sorry for my late reply: I don't have access to the Supermicro hardware or the fork anymore, as I switched jobs last year. Maybe @mxey could add the missing changes to this PR?

mxey:
> Maybe @mxey could add the missing changes to this PR?

Yes, I'll look into this.

oem/supermicro/ikvm.go
mxey: I changed it.

mxey:
@stmcginnis I responded to all the open review comments. I decided to add another commit to the pull request, since previous review changes also appeared to be extra commits. I would suggest squashing it all together at the end.

oem/supermicro/manager.go
stmcginnis: Yes, but it's an unnecessary detail for consumers. In the JSON unmarhalling it can pull out the fields and set them as direct properties of the manager struct.

stmcginnis (commented):
Thanks, I'll take a closer look soon!

oem/supermicro/manager.go
stmcginnis: Yep. You're already using an OEM object, so it feels a little awkward to have to have OEMObject.OEM.Foo instead of OEMObject.Foo.

mxey: We cannot do both. The JSON-based fields have to remain exported for the unmarshalling, so if we inlined `ManagerOem` into `Manager`, all the fields would be visible there. If we leave the extra indirection as-is, we can unexport the `Oem` field, and the JSON fields would all be unexported.

I think we should do it like this:

```
type Manager struct {
	redfish.Manager
	supermicro supermicro
}

type supermicro struct {
	ODataType string `json:"@odata.type"`
	NTP       common.Link
	Syslog    common.Link
	IKVM      common.Link
}

func FromManager(manager *redfish.Manager) (Manager, error) {
	type oemFields struct {
		Supermicro supermicro `json:"Supermicro"`
	}
	var oem oemFields
	err := json.Unmarshal(manager.Oem, &oem)
	if err != nil {
		return Manager{}, fmt.Errorf("can't unmarshal OEM Manager: %v", err)
	}

	return Manager{
		Manager:    *manager,
		supermicro: oem.Supermicro,
	}, nil
}
```

mxey: @stmcginnis what do you think?

oem/supermicro/manager.go
stmcginnis: You need them when unmarshalling, but not after that. See elsewhere in the library where the unmarshalling takes care of accessing internal details, then stores them as non-public fields that are used in a more user-friendly interface.

oem/supermicro/manager.go
stmcginnis: Close, but I think this would be best:

```go
type Manager struct {
	redfish.Manager
	fanMode         string
	iKVM            string
	ipAccessControl string
	kcsInterface    string
	licenseManager  string
	mouseMode       string
	ntp             string
	radius          string
	smcrakp         string
	snmp            string
	smtp            string
	snooping        string
	sysLockdown     string
	syslog          string
}

type managerOem struct {
	Supermicro struct {
		ODataType       string `json:"@odata.type"`
		FanMode         common.Link
		IKVM            common.Link
		IPAccessControl common.Link
		KCSInterface    common.Link
		LicenseManager  common.Link
		MouseMode       common.Link
		NTP             common.Link
		RADIUS          common.Link
		SMCRAKP         common.Link
		SNMP            common.Link
		SMTP            common.Link
		Snooping        common.Link
		SysLockdown     common.Link
		Syslog          common.Link
	} `json:"Supermicro"`
}

func FromManager(manager *redfish.Manager) (Manager, error) {
	var oem managerOem
	err := json.Unmarshal(manager.Oem, &oem)
	if err != nil {
		return Manager{}, fmt.Errorf("can't unmarshal OEM Manager: %v", err)
	}

	return Manager{
		Manager:         *manager,
		fanMode:         oem.Supermicro.FanMode.String(),
		iKVM:            oem.Supermicro.IKVM.String(),
		ipAccessControl: oem.Supermicro.IPAccessControl.String(),
		kcsInterface:    oem.Supermicro.KCSInterface.String(),
		licenseManager:  oem.Supermicro.LicenseManager.String(),
		mouseMode:       oem.Supermicro.MouseMode.String(),
		ntp:             oem.Supermicro.NTP.String(),
		radius:          oem.Supermicro.RADIUS.String(),
		smcrakp:         oem.Supermicro.SMCRAKP.String(),
		snmp:            oem.Supermicro.SNMP.String(),
		smtp:            oem.Supermicro.SMTP.String(),
		snooping:        oem.Supermicro.Snooping.String(),
		sysLockdown:     oem.Supermicro.SysLockdown.String(),
		syslog:          oem.Supermicro.Syslog.String(),
	}, nil
}

func (manager *Manager) NTP() (*NTP, error) {
	return GetNTP(manager.GetClient(), manager.ntp)
}

func (manager *Manager) Syslog() (*Syslog, error) {
	return GetSyslog(manager.GetClient(), manager.syslog)
}

func (manager *Manager) IKVM() (*IKVM, error) {
	return GetIKVM(manager.GetClient(), manager.iKVM)
}
```

mxey: I can do it like that, but it feels very repetitive to me. Why is the type of, for example Manager.fanMode important, when it's not exported anyway?

stmcginnis: You can omit the extra ones that you are not adding accessor methods for. I just put them in for completeness.

This is more repetitive for the implementation side so it's cleaner for the library consumer.

stmcginnis: In other words, we implement it once here so others never need to think about it.

stmcginnis:
Sorry, we've had a push to add more Supermicro support, so unfortunately this is now superceded by other work. There's some great stuff here, so feel free to add to or improve the current code in the repo. Thanks a ton for your effort on this!

PR #214, opened 2022-11-07
client.go: fixes a panic when the client.auth is nil

joelrebel:
The panic occurs if a caller of APIClient attempts to run GetSession() when APIClient.auth has not yet been set.


stmcginnis (approved):
Awesome - thanks Joel!

PR #220, opened 2022-12-12
(think twice) bugfix: huawei EffectiveFamily filed is int

Sn0rt:
We need think twice.

Because the following fields have different types of data from different manufacturers.

Because the following fields have different types of data from different manufacturers.

If it is Huawei, the type is int, if it is dell, it is string

Signed-off-by:  Sn0rt <wangguohao.2009@gmail.com>


Sn0rt:
copy

PR #225, opened 2023-01-30
Fix json.Marshal

Danile71:
Fix  json: unsupported type: func(*http.Request) (*url.URL, error)

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -40,7 +40,7 @@ type Entity struct {
 	// Name is the name of the resource or array element.
 	Name string `json:"Name"`
 	// Client is the REST client interface to the system.
-	Client Client
+	Client Client `json:"-"`
 }
 
 // SetClient sets the API client connection to use for accessing this


stmcginnis:
Thanks @Danile71. Can you tell me under what condition you are hitting this error?

Danile71:
Or need method 

(chassis *Chassis) Marshal() error
```
	chassis, err := service.Chassis()
	if err != nil {
		panic(err)
	}

	for _, chass := range chassis {
		d, err := json.Marshal(chass)
		if err != nil {
			panic(err)
		}


		fmt.Printf(string(d))
	}
```

stmcginnis:
Ah, OK. So it's not part of the marshalling of the object from the REST API response to the GoFish object, but then taking that Gofish object and trying to remarshall it for some reason. Interesting.

stmcginnis (approved):

PR #226, opened 2023-02-14
add oem to AccountService

FalcoSuessgott:
This PR adds the `Oem` to the `AccountService`, in my case needed for setting Global Login Options for Lenovo


stmcginnis (changes requested):
Thanks @FalcoSuessgott, but OEM should be handled a little differently here. Please take a look at the other OEM handlers for how to add a Lenovo OEM object.

https://github.com/stmcginnis/gofish/blob/main/oem/hpe/thermal.go

PR #232, opened 2023-02-27
Add log services to Chassis

xflipped:


stmcginnis (approved):
Thanks @xflipped - looks good to me.

Issue #233, opened 2023-02-28
Tests


Danile71:
tests fail 50/50
maybe problems with parallel?
I will try to see later

```
// complete
PASS
ok  	github.com/stmcginnis/gofish/redfish	0.017s

// fail
=== RUN   TestChassisLinkedDrives
    chassis_test.go:353: Error getting drives: failed to retrieve some items: [{"link":"/redfish/v1/Chassis/HA-RAID.0.StorageEnclosure.0/Drives/Disk.Bay.0","error":"EOF"}]
    chassis_test.go:359: Expected four calls to be made, captured: [{GET /redfish/v1/Chassis/HA-RAID.0.StorageEnclosure.0/Drives/Disk.Bay.0  map[]} {GET /redfish/v1/Chassis/HA-RAID.0.StorageEnclosure.0/Drives/Disk.Bay.1  map[]} {GET /redfish/v1/Chassis/HA-RAID.0.StorageEnclosure.0/Drives/Disk.Bay.3  map[]}]
    chassis_test.go:363: Expected 4 drives to be returned, got 3
--- FAIL: TestChassisLinkedDrives (0.00s)
```

stmcginnis:
Thanks, I did notice one failure, but haven't had a chance to look in to it closer.

PR #234, opened 2023-02-28
Fixe tests and string MaxSpeedMHz

Danile71:
[#223](https://github.com/stmcginnis/gofish/issues/233)


stmcginnis (approved):
Thanks a lot for looking in to this!

Since these are two entirely different things, it would be better to submit them as separate PRs. But good stuff.

PR #236, opened 2023-03-01
fix: add mutex to http client

Danile71:
Our test bmc have problem with parallel requests

random uri with error
`failed to retrieve some items: [{"link":"/redfish/v1/Systems/Self/EthernetInterfaces/EthernetInterface0","error":"403: \u003c?xml version=\"1.0\" encoding=\"iso-8859-1\"?\u003e\n\u003c!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\"\u003e\n\u003chtml xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\"\u003e\n \u003chead\u003e\n  \u003ctitle\u003e403 - Forbidden\u003c/title\u003e\n \u003c/head\u003e\n \u003cbody\u003e\n  \u003ch1\u003e403 - Forbidden\u003c/h1\u003e\n \u003c/body\u003e\n\u003c/html\u003e\n"},{"link":"/redfish/v1/Systems/Self/EthernetInterfaces/EthernetInterface1","error":"403: \u003c?xml version=\"1.0\" encoding=\"iso-8859-1\"?\u003e\n\u003c!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\"\u003e\n\u003chtml xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\"\u003e\n \u003chead\u003e\n  \u003ctitle\u003e403 - Forbidden\u003c/title\u003e\n \u003c/head\u003e\n \u003cbody\u003e\n  \u003ch1\u003e403 - Forbidden\u003c/h1\u003e\n \u003c/body\u003e\n\u003c/html\u003e\n"}]`

### DIFF
--- a/client.go
+++ b/client.go
@@ -416,9 +416,6 @@ func (c *APIClient) RunRawRequestWithHeaders(method, url string, payloadBuffer i
 
 // runRawRequestWithHeaders actually performs the REST calls but allowing custom headers
 func (c *APIClient) runRawRequestWithHeaders(method, url string, payloadBuffer io.ReadSeeker, contentType string, customHeaders map[string]string) (*http.Response, error) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
 	if url == "" {
 		return nil, common.ConstructError(0, []byte("unable to execute request, no target provided"))
 	}
@@ -475,8 +472,9 @@ func (c *APIClient) runRawRequestWithHeaders(method, url string, payloadBuffer i
 			return nil, err
 		}
 	}
-
+	c.mu.Lock()
 	resp, err := c.HTTPClient.Do(req)
+	c.mu.Unlock()
 	if err != nil {
 		return nil, err
 	}
--- a/client.go
+++ b/client.go
@@ -478,12 +478,14 @@ func (c *APIClient) runRawRequestWithHeaders(method, url string, payloadBuffer i
 		}
 	}
 
+	var resp *http.Response
+
 	if c.mu != nil {
 		c.mu.Lock()
-	}
-	resp, err := c.HTTPClient.Do(req)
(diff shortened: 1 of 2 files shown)


Danile71:
@stmcginnis If I parallel get info about managers and systems, and etc.. I have that bug. 
Not only CollectCollection.
In my opinion, bug in bmc firmware.

stmcginnis:
Perhaps we should always use a mutex when sending a request then. It would be safer, and I don't think it would add too much overhead to the call.

client.go
```diff
@@ -410,6 +416,9 @@ func (c *APIClient) RunRawRequestWithHeaders(method, url string, payloadBuffer i
 
 // runRawRequestWithHeaders actually performs the REST calls but allowing custom headers
 func (c *APIClient) runRawRequestWithHeaders(method, url string, payloadBuffer io.ReadSeeker, contentType string, customHeaders map[string]string) (*http.Response, error) {
+	c.mu.Lock()
```
stmcginnis, line 419: Sorry, one more thing while you're at it. Can you lock immediately before the `c.HTTPClient.Do(req)` call, then explicitly unlock right away after? That way we are limiting that window to as small as possible so nothing is blocking while waiting for non-critical things to be done.

Danile71:
@stmcginnis complete

stmcginnis:
Thanks a lot!

PR #237, opened 2023-03-01
fix: SetDefaultBootOrder uri

Danile71:


stmcginnis (approved):
Ouch, that was definitely a mistake. Thanks!

PR #241, opened 2023-03-17
Update event subscription for Redfish v.1.5+

xflipped:
I'd like to suggest some changes for EventService subscriptions

According to [Redfish documentation](https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2022.2.pdf) p.6.34.3

> For the possible property values, see EventTypes in Property details. Deprecated in v1.5 and later. This property has been deprecated. Starting with Redfish Specification v1.6 (Event v1.3), subscriptions are based on the RegistryPrefix and ResourceType properties and not on the EventType property.

Now there's the following error when subscribing to events (Redfish v.1.5+, here: Redfish v.1.15.1):
`400: {"error":{"@Message.ExtendedInfo":[{"@odata.type":"#Message.v1_0_8.Message","Message":"The property EventTypes is not in the list of valid properties for the resource.","MessageArgs":["EventTypes"],"MessageId":"Base.1.12.PropertyUnknown","RelatedProperties":["/EventTypes"],"Resolution":"Remove the unknown property from the request body and resubmit the request if the operation failed.","Severity":"Warning"}],"code":"Base.1.12.PropertyUnknown","message":"The property EventTypes is not in the list of valid properties for the resource."}}`

The changes I've made are backward-compatible

### DIFF
--- a/redfish/eventdestination.go
+++ b/redfish/eventdestination.go
@@ -272,20 +272,50 @@ func GetEventDestination(c common.Client, uri string) (*EventDestination, error)
 
 // subscriptionPayload is the payload to create the event subscription
 type subscriptionPayload struct {
-	Destination string                   `json:"Destination"`
-	EventTypes  []EventType              `json:"EventTypes"`
-	HTTPHeaders map[string]string        `json:"HttpHeaders,omitempty"`
-	Oem         interface{}              `json:"Oem,omitempty"`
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):
Awesome, thanks @xflipped!

I'm sure there are more recent protocol updates that need to be picked up. This is probably one of the more important ones.

PR #242, opened 2023-03-29
Add network protocols support

kozl:
Hi! Thank you very much for this library. I needed support of the NetworkProtocol resource of Manager, so here is my contribution. 
NetworkProtocolSettings struct is based on [this schema](http://redfish.dmtf.org/schemas/v1/ManagerNetworkProtocol.v1_9_1.yaml)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,7 +7,7 @@ linters-settings:
   goheader:
     values:
       const:
-       - LICENSE: BSD-3-Clause
+        - LICENSE: BSD-3-Clause
     template: |-
       SPDX-License-Identifier: {{ LICENSE }}
   goconst:
@@ -95,9 +95,9 @@ linters:
 issues:
   exclude:
     - 'declaration of "(err|ctx)" shadows declaration at'
-    - 'G402: TLS InsecureSkipVerify may be true'
-    - '`(F|f)ibre` is a misspelling'
-    - '`(C|c)ancell.*` is a misspelling'
+    - "G402: TLS InsecureSkipVerify may be true"
+    - "`(F|f)ibre` is a misspelling"
+    - "`(C|c)ancell.*` is a misspelling"
 run:
   skip-dirs:
     - examples
--- a/common/entity_test.go
+++ b/common/entity_test.go
@@ -22,7 +22,7 @@ type testStruct struct {
 	E nestedStruct
 }
 
-func TestGetPatchPayloadFromUpdate(t *testing.T) {
+func TestGetPatchPayloadFromUpdate(t *testing.T) { //nolint:funlen
 	testcases := []struct {
 		name           string
 		originalEntity interface{}
--- a/common/entity_test.go
+++ b/common/entity_test.go
@@ -1,3 +1,7 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+//
+
 package common
 
 import (
--- a/common/entity_test.go
+++ b/common/entity_test.go
@@ -1,0 +1,205 @@
+package common
+
+import (
+	"reflect"
+	"testing"
+)
+
+type nestedStruct struct {
+	X bool `json:"x"`
+	Y []string
+}
+
+type testStruct struct {
+	A string `json:"a"`
+	B []string
+	C map[string]string
+	D int
+	E nestedStruct
+}
+
+func TestGetPatchPayloadFromUpdate(t *testing.T) {
+	testcases := []struct {
+		name           string
+		originalEntity interface{}
+		updatedEntity  interface{}
+		expected       map[string]interface{}
+	}{
+		{
+			name: "no updates were made",
+			originalEntity: &testStruct{
+				A: "abcd",
+				B: []string{"A", "B", "C", "D"},
+				C: map[string]string{
+					"A": "a",
+					"B": "b",
+					"C": "c",
+					"D": "d",
+				},
+				D: 1,
+			},
+			updatedEntity: &testStruct{
+				A: "abcd",
+				B: []string{"A", "B", "C", "D"},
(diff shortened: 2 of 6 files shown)


kozl:
Is there support for updating fields like NetworkProtocolSettings.NTP.NTPServers ? I saw your comment [here](https://github.com/stmcginnis/gofish/blob/main/common/entity.go#L46). Looks like it's not, am I right?

stmcginnis:
This is looking great!

Yeah, we might need to try out the `Update` handling with something like this and see if anything needs to be done. It might be able to just pass the array of values, but I haven't had a chance to look too closely and can't recall off the top of my head.

kozl:
Ok, I'll try to add structures and arrays support in Update, I think today if it won't be very hard. Hope you could release this changes as soon as you can, I need network protocol support in my project. Thank you!

kozl:
Ok, I'll try to add structures and arrays support in Update, I think today if it won't be very hard. Hope you could release this changes as soon as you can, I need network protocol support in my project. Thank you!

kozl:
Ok, I'm done. I've put the payload logic in a separate funciton, write tests on it and add support for updating slices and structs.

go.sum
```diff
@@ -0,0 +1,16 @@
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
stmcginnis, line 1: testify is great, but one goal I've been trying to adhere to is to not pull in any other dependencies unless they are really necessary. Can you use the standard library functionality for this?

kozl: Ok, done

stmcginnis: Awesome, thanks!

kozl:
Fixed failed tests (global strings.Reader was read twice)

stmcginnis:
Still a few things to address from the earlier job failures. Will wait to approve the workflow run for that.

Thanks for the updates!

kozl:
Is there something I need to do, to merge this?

stmcginnis:
> Still a few things to address from the earlier job failures. Will wait to approve the workflow run for that.
> 
> Thanks for the updates!

Yep, there were a few failures that haven't been addressed yet. Once those are resolved, this looks good! Running `make` locally will run the same jobs as the GitHub Actions.

Sorry, looks like there is also a merge conflict now, so you will also need to rebase the latest from the `main` branch.

README.md
```diff
@@ -2,8 +2,8 @@
 
 [![Go Doc](https://godoc.org/github.com/stmcginnis/gofish?status.svg)](http://godoc.org/github.com/stmcginnis/gofish)
 [![Go Report Card](https://goreportcard.com/badge/github.com/stmcginnis/gofish?branch=main)](https://goreportcard.com/report/github.com/stmcginnis/gofish)
-[![Releases](https://img.shields.io/github/release/stmcginnis/gofish/all.svg?style=flat-square)](https://github.com/stmcginnis/gofish/releases)
-[![LICENSE](https://img.shields.io/github/license/stmcginnis/gofish.svg?style=flat-square)](https://github.com/stmcginnis/gofish/blob/main/LICENSE)
+[![Releases](https://img.shields.io/github/release/stmcginnis/gofish/all.svg?style=flat-square)](https://github.com/kozl/gofish/releases)
+[![LICENSE](https://img.shields.io/github/license/stmcginnis/gofish.svg?style=flat-square)](https://github.com/kozl/gofish/blob/main/LICENSE)
```
stmcginnis, line 6: I can enable the workflow to run, but first I think you made a bunch of these changes by accident. I usually just use a replace directive if I'm testing out local changes to a library so there's less mucking with the real source code.

There are a few lint issues from the earlier linting check, but doesn't look like it kept those comments around after the updates.

kozl: Fixed this

kozl:
Could we merge this now? Anything else left?

common/entity_test.go
```diff
@@ -0,0 +1,205 @@
+package common
```
stmcginnis, line 1: Missing license header. I remember that being one of the linter errors.

redfish/network_protocol.go
```diff
@@ -0,0 +1,292 @@
...
+	NoneEncryption SNMPEncryptionProtocol = "None"
+	// Encryption is determined by account settings
+	AccountEncryption SNMPEncryptionProtocol = "Account"
+	// CBC-DES encryption
+	CBC_DES_Encryption SNMPEncryptionProtocol = "CBC_DES" // nolint
+	// CFB128-AES-128 encryption
+	CFB128_AES128_Encryption SNMPEncryptionProtocol = "CFB128_AES128" // nolint
+)
+
+type EngineId struct {
```
stmcginnis, line 107: Should be `EngineID`.

As well as all the other `*Id` names in here.

kozl: Done!

stmcginnis:
Almost there!
image

kozl:
Fixed nolint comment issues and added exclude-rules part in golang-ci lint (I suppose we don't want to check the length of testing functions)

.golangci.yaml
```diff
@@ -7,7 +7,7 @@ linters-settings:
   goheader:
     values:
       const:
-       - LICENSE: BSD-3-Clause
+        - LICENSE: BSD-3-Clause
```
stmcginnis, line 10: Did you mean to include these changes? They should not be needed.

stmcginnis: Nah, if some yaml auto-formatter does this, chances are it will just happen again in the future. Doesn't hurt to change it, so let's keep it.

stmcginnis (approved):
Nice work - thanks!

PR #243, opened 2023-03-31
Get log entries with pagination

xflipped:
I've faced with the following situation when retrieving logs from LogService:

Link to entries:
```
  "@odata.type": "#LogService.v1_1_3.LogService",
  "Entries": {
    "@odata.id": "/redfish/v1/Managers/{ManagerID}/LogServices/{LogServiceID}/Entries"
  }
```
Actual entry collection object:
```
  "Members@odata.count": 52,
  "Members@odata.nextLink": "/redfish/v1/Managers/{ManagerID}/LogServices/{LogServiceID}/Entries?$skip=50",
  "Name": "Log Service Entries Collection"
```

As you can see, this is the case when the original link leads only to the first N (here: N=50) entries. So I would like to propose changes that allow all entries to be retrieved.

I've also noticed that even though the log entry collection has a json field "Members" that normally contains links, here it contains a list of log entries directly:

Normally:
```
  "Members": [
    {
      "@odata.id": "/redfish/v1/Managers/{ManagerID}"
    }
  ],
```
Logs:
```
"Members": [
{
"@odata.id":view details "/redfish/v1/Managers/{ManagerID}/LogServices/{LogServiceID}/Entries/1",
"@odata.type":view details "#LogEntry.v1_14_0.LogEntry",
"Id":"1"
},
] 
```

So I think there is no need to query each log entry as a link. These changes can also be seen in my PR


stmcginnis:
> I've also noticed that even though the log entry collection has a json field "Members" that normally contains links, here it contains a list of log entries directly

This vendor is not following the Redfish specification:

```json
"Entries": {
    "$ref": "http://redfish.dmtf.org/schemas/v1/LogEntryCollection.json#/definitions/LogEntryCollection",
    "description": "The link to the log entry collection.",
    "longDescription": "This property shall contain a link to a resource collection of type LogEntryCollection.",
    "readonly": true
},
```

An issue should be opened with that vendor asking that they fix their implementation to follow the spec.

redfish/logentrycollection.go
```diff
@@ -0,0 +1,59 @@
...
+	// Description provides a description of this resource.
+	Description string
+	// Members contain log entries of this log entry collection
+	Members []*LogEntry `json:"Members"`
+	// MembersCount is the number of log entries of this log entry collection.
+	MembersCount int `json:"Members@odata.count"`
+	// MembersNextLink is the link used for pagination
+	MembersNextLink string `json:"Members@odata.nextLink"`
+	// rawData holds the original serialized JSON so we can compare updates.
+	rawData []byte
```
stmcginnis, line 30: `rawData` would not be needed here. That's only necessary when a structure has updatable fields where an `Update()` call is required.

redfish/logentrycollection.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// UnmarshalJSON unmarshals a LogEntry object from the raw JSON.
+func (logEntryCollection *LogEntryCollection) UnmarshalJSON(b []byte) (err error) {
```
stmcginnis, line 34: Since `rawData` isn't needed to be captured, this custom marshalling can be removed completely.

stmcginnis:
And the `LogEntryCollection.Members` defintion:

```json
"Members": {
    "autoExpand": true,
    "description": "The members of this collection.",
    "items": {
        "$ref": "http://redfish.dmtf.org/schemas/v1/LogEntry.json#/definitions/LogEntry"
    },
    "longDescription": "This property shall contain an array of links to the members of this collection.",
    "readonly": true,
    "type": "array"
},
```

stmcginnis (changes requested):
Looks close, and good overall. A few small things, and we need to make sure it is following the actual spec and not one vendors flawed implementation.

Thanks for working on this!

xflipped:
Thanks for your comments!! 

As for this
`An issue should be opened with that vendor asking that they fix their implementation to follow the spec.`

Please let me check this issue because I took the example directly from the dmtf website here: 
https://redfish.dmtf.org/redfish/mockups/v1/1266#Systems--437XR1138R2--LogServices--Log1--Entries

(https://redfish.dmtf.org/redfish/mockups/v1/1266  -> Systems > 437XR1138R2 > LogServices >Log1 > Entries)

xflipped:
I've added filtering as an option and vendor agnostic collection retrieval using links to follow the spec

stmcginnis (approved):
Thanks @xflipped!

GlqEason:
so it still  need to query each log entry as a link ?  i have tested some manufacturer device, and /redfish/v1/Managers/{ManagerId}/LogServices/{LogServiceId}/Entries api can got all enties completely

GlqEason:
> there's no need to query each log entry as a link, a collection will be retrieved correctly taking into account pagination and as an option you can apply filtering (top, skip)

sorry, i still don't quite understand, and which function can i use to get the latest 10 log entries，at least i should know how many logs it has ?

PR #244, opened 2023-04-04
fix: set client private with getter/setter

Danile71:
`Client Client `json:"-"``

Fix only for json, but bson, etc...


stmcginnis (approved):
That works.

PR #248, opened 2023-04-12
fixed bug with Update for embedded structs

kozl:
Found out that if you update embedded struct field you got wrong payload to patch. Fixing it


stmcginnis (approved):
Thanks for the quick fix.

Issue #246, opened 2023-04-07
ComputerSystem.Reset add response data to function return


sseekamp:
Here: https://github.com/stmcginnis/gofish/blob/da46c1326ae9ca2e51f4c7482e6eae4bbf2e1c75/redfish/computersystem.go#L850-L875

With many Redfish implementations now returning additional data (task urls specifically) it would be appreciated if the function above (and probably others) also returned the response or select attributes of it to the caller.

stmcginnis:
Thanks, this looks like it would be good to expand on these types of actions. At least in the case of `ComputerSystem.Reset`, the response could/should look at the response code and if it is `202` handle returning a task:

```yaml
/redfish/v1/Systems/{ComputerSystemId}/Actions/ComputerSystem.Reset:
  post:
    parameters:
    - description: The value of the Id property of the ComputerSystem resource
      in: path
      name: ComputerSystemId
      required: true
        type: string
    requestBody:
      content:
        application/json:
          schema:
            $ref: http://redfish.dmtf.org/schemas/v1/ComputerSystem.v1_6_0.yaml#/components/schemas/ResetRequestBody
      required: true
    responses:
      '200':
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/RedfishError'
        description: The response contains the results of the Reset action
      '202':
        content:
          application/json:
            schema:
              $ref: http://redfish.dmtf.org/schemas/v1/Task.v1_4_0.yaml#/components/schemas/Task
        description: Accepted; a task has been generated
      '204':
        description: Success, but no response data
      default:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/RedfishError'
        description: Error condition
```

sseekamp:
Excellent!

Before I get in over my head with trying to implement that, would you suggest using a common.Operations object or a Task object or just return the raw URL string with the task link if a 202?

stmcginnis:
I think it should probably return a `Task` object to stay somewhat consistent with the spec definition. I've tried to avoid returning a raw URL string as much as possible to hide the low level details.

The tricky part of this is going to be that the [current `Post` method only returns an `error`](https://github.com/stmcginnis/gofish/blob/8552e3743ac735840b38ec31df99a709dc46344c/common/entity.go#L104). Since there are several operations that can return a `Task`, it probably makes sense to change this to return a tuple of `(Task, error)`.

[...]

PR #249, opened 2023-04-12
ComputerSystemReset: Refactor Task and Post

sseekamp:
Possible direction for: https://github.com/stmcginnis/gofish/issues/246

- Moves Task into common package
- Change ComputerSystem.Reset() to return a *common.Task struct
- Change ComputerSystem.Post() to return an *http.Response (which is one level additional from PostWithHeaders)
- Updated method calls in other locations to accommodate new signatures.

### DIFF
--- a/common/collection.go
+++ b/common/collection.go
@@ -118,7 +118,7 @@ func CollectList(get func(string), c Client, link string) error {
 	return nil
 }
 
-// CollectCollection will retrieve a collection of entitied from the Redfish service
+// CollectCollection will retrieve a collection of entities from the Redfish service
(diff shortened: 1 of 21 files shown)


stmcginnis:
A few linting issues that need to be addressed, and now there are merge conflicts with a few of the files. Can you update?

Thanks!

stmcginnis:
Thanks a lot for looking in to this. There are some other changes that conflict with this, and there hasn't been a response for fixing the linting issues, so I'm going to close this for now. If you have the desire to continue with these changes then feel free to get an updated PR up and we can discuss. Thanks!

PR #254, opened 2023-06-12
Add sensors located in the chassis and sub-components. (v1.9+)

xflipped:
Implemented according to
https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2023.1.html#sensor-170


stmcginnis (approved):
Thanks!

PR #256, opened 2023-06-14
Add serial interfaces in manager

xflipped:
Implemented according to https://redfish.dmtf.org/schemas/v1/SerialInterface.v1_1_8.yaml


stmcginnis (approved):
Looks good!

PR #257, opened 2023-06-15
Added job service

rizwan-namutech:


redfish/job.go
```diff
@@ -0,0 +1,131 @@
+package redfish
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/stmcginnis/gofish/common"
```
stmcginnis, line 6: This will need to be moved down into it's own "local" import group to follow the linter rules.

redfish/jobservice.go
```diff
@@ -0,0 +1,129 @@
...
+	ODataContext string `json:"@odata.context"`
+	// ODataType is the odata type.
+	ODataType string `json:"@odata.type"`
+	// Description provides a description of this resource.
+	Description string
+	// DateTime system time.
+	DateTime time.Time
+	// jobs points towards the jobs collection
+	jobs string
+	// jobs points towards the jobs collection
```
stmcginnis, line 33: s/jobs/logs/

redfish/job.go
```diff
@@ -0,0 +1,131 @@
+package redfish
```
stmcginnis, line 1: The new files will need the SPDX license header added.

redfish/jobservice.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Extract the links to other entities for later
+	*jobService = JobService(t.temp)
+	jobService.jobs = t.Jobs.String()
+
+	return nil
+}
+
+// DeleteJob deletes a specific job using the job service.
+func (jobService *JobService) DeleteJob(uri string) error {
```
stmcginnis, line 121: This should probably be changed to a method on the `Job` object itself so someone can just call `job.Delete()`. There are a few cases where it can't be avoided, but in general we should try to avoid cases where the consumer needs to handle URI's directly.

stmcginnis (changes requested):
A few minor things, but this looks great. Thanks for working on it!

redfish/jobservice.go
rizwan-namutech: Fixed

rizwan-namutech:
I have fixed the linting issues.

riethm (commented):
I was preparing a very similar PR, so you've saved me some work here.

In comparing branches, I noted a few minor inconsistencies with other service implementations that I thought I'd bring it to your attention. They're entirely cosmetic, and otherwise (for what it's worth) this looks great.

stmcginnis:
Sorry I got tied up with some other things and didn't get back to this for so long. There are some interesting things in https://github.com/stmcginnis/gofish/pull/266. Maybe we can reconcile the work between the two?

stmcginnis:
Sorry for not getting back to this right away. Decided to go ahead with #266, but I noted you as co-author in the squash commit. I'm going to close this PR now. Thanks for working on it!

PR #261, opened 2023-07-04
make concurrency configurable

jpiper:
This allows the concurrency of HTTP client to be configured (it is 1 by default in order to maintain the existing behaviour). This also has the added benefit that a context cancellation will abort the lock acquisition

### DIFF
--- a/client.go
+++ b/client.go
@@ -22,8 +22,6 @@ import (
 	"strings"
 	"time"
 
-	"golang.org/x/sync/semaphore"
-
 	"github.com/stmcginnis/gofish/common"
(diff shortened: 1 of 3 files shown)


stmcginnis:
Thanks! Any way we can accomplish this without using `golang.org/x/sync`? I know that library has had some security issues in the past, and I've been trying really hard to keep gofish from needing to pull in any extra dependencies.

jpiper:
yes I can rewrite this using a simple channel, I'll do that now for you

jpiper:
I think that does the trick

stmcginnis (approved):
Thanks @jpiper!!

PR #266, opened 2023-07-27
Add jobservice and job types

riethm:
Alternative to #257 (abandoned?) that addresses issues noted in that PR:
* Rename `ServiceCapabilities` to `JobServiceCapabilities` (consistent with schema)
* Rename service capability `ServiceEnabled` to `Scheduling` (consistent with schema. `ServiceEnabled` is a property of `JobService` rather than a capability)
* export `ListReferencedJobs` (consistent with other gofish library services)
* Move `GetJob` and `ListReferencedJobs` from jobservice.go to job.go (consistent with other gofish library services)


stmcginnis (approved):
Thanks for the patience, I wanted to see if we had a response on the other related PR. Let's go ahead with this one and see how it goes.

PR #268, opened 2023-08-03
fix: Pass payload during bios reset

kourtneyboss:
ResetBios fails with a nil payload. An empty payload is accepted.


stmcginnis (approved):
This looks fine to me, and no one has come along raising any issues. The chance of that happening might be better if it's actually part of the codebase, so let's merge it and watch for anyone having problems.

I don't personally have the hardware to test this against a variety of systems, so unfortunately we will have to use the real world as a test. ;)

PR #271, opened 2023-08-30
Add HttpBootUri into Boot

lzap:
This flag allows to override UEFI HTTP Boot URI, only works when BootSourceOverrideTarget is set to UefiHttp.

I could not find any unit test for updating boot structure, let me know if I missed it. Thanks!

### DIFF
--- a/redfish/computersystem.go
+++ b/redfish/computersystem.go
@@ -328,6 +328,8 @@ type Boot struct {
 	UefiTargetBootSourceOverride string `json:",omitempty"`
 	// The link to a collection of certificates used for booting through HTTPS by this computer system.
 	certificates string
+	// The URI to boot from when BootSourceOverrideTarget is set to UefiHttp.
+	HTTPBootURI string `json:",omitempty"`
 }
 
 // UnmarshalJSON unmarshals a Boot object from the raw JSON.
--- a/redfish/computersystem_test.go
+++ b/redfish/computersystem_test.go
@@ -59,7 +59,8 @@ var computerSystemBody = `{
 				"SDCard",
 				"UefiHttp"
 			],
-			"UefiTargetBootSourceOverride": "uefi device path"
+			"UefiTargetBootSourceOverride": "uefi device path",
+			"HttpBootURI": "http://localhost/boot.efi"
 		},
 		"BiosVersion": "P79 v1.00 (09/20/2013)",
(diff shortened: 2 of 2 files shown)


stmcginnis:
There's a minor linting issue. Can you address that?

lzap:
Absolutely, force-pushed.

lzap:
So I tested on a DELL R350 and it works, unfortunately this system does not support this Redfish field yet :-D

stmcginnis:
> So I tested on a DELL R350 and it works, unfortunately this system does not support this Redfish field yet :-D

Oh well, I guess we at least know it doesn't make older systems blow up. :D

stmcginnis:
> Absolutely, force-pushed.

Shoot, part way there! Linter still not happy yet.

lzap:
Yeah I was waiting until it passes before I make another comment, sorry about that :-)

Thanks. FYI I am building a small prototype, a provisioning service for Red Hat systems, it’s an image-based provisioning via Anaconda and HTTP UEFI Boot via Redfish, fully automated:

https://github.com/foresterorg/forester

Thanks again. Cheers.

lzap:
I finally got to hardware testing this feature and I noticed that the field is serialized as `HTTPBootURI`. Linter made me to change the name, I am not sure if Redfish API is actually case-insensitive so it would not make any difference. Currently I loaned 6 systems from Dell and SuperMicro and any of them does not support this relatively new feature so I cannot confirm for sure :-(

stmcginnis:
> I finally got to hardware testing this feature and I noticed that the field is serialized as `HTTPBootURI` whereas the spec says `HttpBootUri`. Linter made me to change the name, I am not sure if Redfish API is actually case-insensitive so it would not make any difference. Currently I loaned 6 systems from Dell and SuperMicro and any of them does not support this relatively new feature so I cannot confirm for sure :-(

If/when you are able to test, and if the casing is an issue, that is handled by giving a serialize/deserialize name to the JSON settings. An example can be seen [here](https://github.com/stmcginnis/gofish/blob/main/redfish/logentry.go#L344).

lzap:
I still don’t have access to any system that supports this, from what I have learned modern HPE systems do support it. Anyway, I want to fix this anyway: https://github.com/stmcginnis/gofish/pull/292

PR #275, opened 2023-09-15
Add option to clean etag before PATCH and POST

BruceNocentini13:
While using this client with Intel machines, we came across an issue where PATCH and POST requests were failing due to unmatched etags.
Some vendors surround header etag with quotes, which need to be cleaned before sending a patch.

This MR simply adds an option to trim quotes from the etag before it is added to the header of the request

### DIFF
--- a/common/entity.go
+++ b/common/entity.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 )
 
 // Entity provides the common basis for all Redfish and Swordfish objects.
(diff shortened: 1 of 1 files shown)


stmcginnis:
Any idea if it's safe to just always do this?

BruceNocentini13:
Don't think it's safe to always have it enabled. This issue only applies to some vendors. In our case we came across it when dealing with some Intel machines. That's why I added a flag.
A second option is to simply not add the etag to the header using "If-Match", but that would essentially remove support for etags

stmcginnis (approved):
Seems reasonable - thanks!

PR #277, opened 2023-09-20
common/entity: adds parameter to skip the If-Match Etag header

joelrebel:
This is a work around for (crappy) vendor implementations where the `If-Match` header does not work - even with '*' as the value, requests are incorrectly denied with an ETag mismatch error.


stmcginnis (approved):
Thanks Joel!

PR #278, opened 2023-09-28
Add new fields to processor

xflipped:
Implemented according to
https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2023.1.html#processor-1180


stmcginnis (approved):
Looks great!

PR #280, opened 2023-10-02
feat(manageraccount): add oem field

FalcoSuessgott:


FalcoSuessgott:
Changed it to `json.RawMessage`. I could also change the remaining fields as well, in a separate commit, if you want.

Added a test for the `oem` attribute, does that work for you? Haven't found any example test assertions for oem fields.

stmcginnis (approved):
Thanks!

PR #289, opened 2023-11-21
redfish/task: Adds the Oem field as json.RawMessage

joelrebel:
This exposes the Task `Oem` field to the caller, its upto the caller to unmarshal its contents.

@stmcginnis  Let me know if this should go into `common.Entity` or if it should be done differently.


stmcginnis (approved):
This looks good, thanks @joelrebel

PR #297, opened 2024-01-23
Fixed the issue: "session were not getting deleted in redfish"

imanubhav:


stmcginnis (changes requested):
Thanks for looking into this @imanubhav. But in general, we do not want library code printing to stdout. Maybe it would be better to have the logout method bubble up the error and let the consumer handle that if they want to.

We also want to avoid returning internal details like the URI called. That is useful for debugging, but consumers of the gofish library shouldn't need to know those internal details. Maybe add it as an optional tracing functionality that can be enabled if/when needed?

redfish/session.go
```diff
@@ -6,7 +6,6 @@ package redfish
 
 import (
 	"net/url"
-
```
stmcginnis, line 9: There still needs to be this separation between the std library import group and local imports.

redfish/session.go
```diff
@@ -103,7 +102,7 @@ func CreateSession(c common.Client, uri, username, password string) (auth *AuthT
 
 // DeleteSession deletes a session using the location as argument
 func DeleteSession(c common.Client, sessionURL string) (err error) {
-	resp, err := c.Delete(sessionURL)
+	resp, err := c.Delete("/redfish/v1/SessionService/Sessions/" + sessionURL)
```
stmcginnis, line 105: Can you confirm - is the issue here from `client.go` where `DeleteSession` is being called, that it is only passing through the session ID rather than the full URL path for the session? I haven't been able to test with a real system.

If that is the case, please rename the parameter variable name to `sessionID` to make it clear what is expected to be passed in to this call, as well as in `serviceroot.go`.

Thanks!

stmcginnis:
Thanks for looking in to this. I got a chance to test this some more. Under normal usage, this change is not correct. The session should contain the full URI:

Logout

If there is some other condition under normal usage that would lead to this not being correct then I would need more details to understand what that is. We would also need to handle the unlikely but possible scenario where that SessionService path is not the one being hard coded here.

PR #305, opened 2024-03-06
Add LocationIndicatorActive as a field in chassis

### LINKED ISSUE #304: add LocationIndicatorActive field to chassis.go
IndicatorLED was deprecated in v1.14 in favor of `LocationIndicatorActive`

It's currently not a field in `redfish/chassis.go` but has been added in other files.

iamsli:
Closes #304

### DIFF
--- a/redfish/chassis.go
+++ b/redfish/chassis.go
@@ -183,6 +183,7 @@ type Chassis struct {
 	HeightMm float64
 	// IndicatorLED shall contain the indicator light state
 	// for the indicator light associated with this system.
+	// Deprecated v1.14+ in favor of LocationIndicatorActive property
 	IndicatorLED common.IndicatorLED
 	// Location shall contain location information of the
 	// associated chassis.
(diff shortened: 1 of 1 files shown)


redfish/chassis.go
```diff
@@ -187,6 +187,8 @@ type Chassis struct {
 	// Location shall contain location information of the
 	// associated chassis.
 	Location common.Location
+	// An indicator allowing an operator to physically locate this resource. (v1.14+)
```
stmcginnis, line 190: Could you add a note to the comment for [IndicatorLED](https://github.com/stmcginnis/gofish/blob/main/redfish/chassis.go#L185) noting that it is deprecated and this field should be checked instead?

stmcginnis (changes requested):
One quick update, if you wouldn't mind. Thanks for looking into this!

stmcginnis (approved):

Issue #322, opened 2024-04-30
Error getting Chassis collection for DPU BMCs


iamsli:
Getting the below error when trying to get the chassis collection by 
`serviceRoot.Chassis()` on DPU BMCs 

```
error getting list of chassis: failed to retrieve some items: [{"link":"/redfish/v1/Chassis/Card1","error":"json: cannot unmarshal object into Go struct field .Controls of type common.Links"},{"link":"/redfish/v1/Chassis/Bluefield_BMC","error":"json: cannot unmarshal object into Go struct field .Controls of type common.Links"}]chassisList: [0xc0001c5b08]
```

stmcginnis:
Ah, I think I see what the issue is. I can have a fix up shortly.

I'll start working on that, but just to confirm, would you be able to run the following to get the output from this machine? I've never worked with a DPU machine, so it would be great to see what it returns.

```sh
curl https:/$ADDR/redfish/v1/Chassis/Bluefield_BMC \
    -X GET \
    -k -s \
    -H "Content-type: application/json" \
    -u "$USER:$PASS" | jq .
```

If you don't have `jq` available you can omit that and just get the raw JSON response.

If you are unable to use `curl`, no worries. If you do know a way to get the output from the system with the tools you have, great. If not, I'm 99% certain I know the problem

PR #323, opened 2024-04-30
make getting Chassis collection work with DPU BMCs

iamsli:
This fixes the issue where we cannot get the Chassis collection for DPU BMCs. 

Per Redfish docs, Controls should be a `Link` singular

```
Controls (v1.17+) { object The link to the collection of controls located in this chassis.
Contains a link to a resource.
@odata.id string read-only Link to Collection of Control. See the Control schema for
details.
}
```

Closes: #322

PR #324, opened 2024-04-30
Fix Chassis controls link definition

stmcginnis:
The Controls property of the Chassis object was incorrectly defined as a `common.Links` type, but it is actually a link to a collection, so it should be a `common.Link` (singular). This changes it to the Link type to avoid a JSON marshalling error on systems that populate this property.

This also uncovered that the link that is used later to retrieve these controls was never actually being set. Both issues are now fixed.

Fixes: #322


stmcginnis:
Being handled in #323

PR #327, opened 2024-05-05
Add AssociatedMACAddresses field to Port

smiller248:


stmcginnis:
Thanks @smiller248. But `AssociatedMACAddresses` is actually [part of](https://github.com/stmcginnis/gofish/blob/main/redfish/port.go#L697) the [`EthernetProperties` struct](https://github.com/stmcginnis/gofish/blob/main/redfish/port.go#L263), so this should be `Port.Ethernet.AssociatedMACAddresses`. You can find the current schema for `Port` [here](https://redfish.dmtf.org/schemas/v1/Port.v1_11_0.json).

smiller248:
Ah, I was confused. Thanks for pointing me in the right direction. Closing this.

PR #329, opened 2024-05-06
Swtich NetworkAdapter fields from common.LinksCollection to common.Link

smiller248:
This changes a number of properties of the `NetworkAdapter` resource type (`NetworkPorts`, `Ports`, `Processors`, etc.) to be parsed as `common.Link` instead of `common.LinksCollection`. Meaning that each of these properties will now be parsed a single link to a collection resource rather than as an array of links.

(For `NetworkDeviceFunctions` and `NetworkPorts`, the flip from `common.LinksCollection` to `common.Link` reverses the change that occurred in https://github.com/stmcginnis/gofish/commit/646d2abd848f96887a93d47e5d148294664e6572. The other properties were newly added to gofish.)

This is necessary for`NewtworkAdapter` to parse correctly. Please correct me if I'm somehow confused here, but the spec for the latest `NetworkAdapter` schema (https://redfish.dmtf.org/schemas/v1/NetworkAdapter.v1_10_0.json) states that all of these fields are single links, not arrays. (I think there may have been some confusion due to`ControllerLinks` having child properties with the same names that are in fact arrays, not single links, e.g. there is `ControllerLinks.NetworkPorts` which is an array).

### DIFF
--- a/redfish/networkadapter.go
+++ b/redfish/networkadapter.go
@@ -317,7 +317,7 @@ type NetworkAdapter struct {
 	assembly string
 	// Certificates shall contain a link to a resource collection of type CertificateCollection that contains
 	// certificates for device identity and attestation.
-	certificates []string
+	certificates string
 	// Controllers shall contain the set of network controllers ASICs that make
 	// up this NetworkAdapter.
 	Controllers []Controllers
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks! There were definitely a few spots where this error was made when adding a lot of new objects in the last big update. Thanks for catching that.

It would be great to extend the unit test JSON to include these to make sure everything parses correctly, but I've had to fix up a few like this elsewhere so I'm pretty confident this is good.

PR #345, opened 2024-05-23
expose the ActionInfo property under Actions under the Managers colle…

iamsli:
exposes the ActionInfo property under Actions under the Managers collection

per Redfish docs `ActionInfo` is a common property across the `Actions` object. Exposing that property in the Managers collection

### DIFF
--- a/redfish/manager.go
+++ b/redfish/manager.go
@@ -379,6 +379,8 @@ type Manager struct {
 	modifyRedundancySetTarget string
 	// resetTarget is the internal URL to send reset targets to.
 	resetTarget string
+	// resetInfo contains URI for an ActionInfo Resource that describes this action.
+	actionInfo string
 	// SupportedResetTypes, if provided, is the reset types this system supports.
 	SupportedResetTypes   []ResetType
 	resetToDefaultsTarget string
@@ -393,6 +395,7 @@ func (manager *Manager) UnmarshalJSON(b []byte) error {
 		ForceFailover       common.ActionTarget `json:"#Manager.ForceFailover"`
 		ModifyRedundancySet common.ActionTarget `json:"#Manager.ModifyRedundancySet"`
 		Reset               struct {
+			ActionInfo        string      `json:"@Redfish.ActionInfo"`
 			AllowedResetTypes []ResetType `json:"ResetType@Redfish.AllowableValues"`
 			Target            string
 		} `json:"#Manager.Reset"`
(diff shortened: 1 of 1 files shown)


stmcginnis:
Thanks. It's a recommended property, but from what I've seen there are only a small number of vendors that actually implement it so far.

This just captures the string to a private property. Is there more you are looking at adding to do something with this? It may be better to add it to the [ActionInfo](https://github.com/stmcginnis/gofish/blob/main/common/types.go#L698) struct and update this spot to use a struct that extends that to get the `AllowedResetTypes`.

iamsli:
ah thank you. i didn't see that struct. i will update this now. for reference this is the output i'm seeing:

```
  "Actions": {
    "#Manager.Reset": {
      "@Redfish.ActionInfo": "/redfish/v1/Managers/Bluefield_BMC/ResetActionInfo",
      "target": "/redfish/v1/Managers/Bluefield_BMC/Actions/Manager.Reset"
    },
```

a call to manager.Reset(<RESET_TYPE>) is failing because nothing is reported under `allowableValues` even though the vendor has this implemented, but to find this info we have to get it from the `ActionInfo` endpoint. How do you think this should be handled?

stmcginnis:
> a call to manager.Reset(<RESET_TYPE>) is failing because nothing is reported under allowableValues

That wouldn't be why the reset call is failing. If there are no reported allowable reset types, [it will just skip performing the check up front](https://github.com/stmcginnis/gofish/blob/main/redfish/manager.go#L586). So the type of reset you are requesting must not be one of the supported kinds, but that wouldn't change the result you are getting. It's just a little more efficient for Gofish to tell you right away if something isn't supported, versus attempting to perform the action and the system telling you the same thing.

So you most likely just need to change the `ResetType` that you are passing in. Gofish would just give you a [slightly better failure message](https://github.com/stmcginnis/gofish/blob/main/redfish/manager.go#L600) maybe, but it still wouldn't tell you what the valid values would be.

iamsli:
correct me if i'm wrong but the gofish `Reset()` makes a Reset call with an empty payload when `AllowableValues` is empty. In this case, AllowableValues is empty so i'm are making an empty call and getting an error message. The reset type i'm passing in via gofish and via curl are the same. the curl call is working but not the gofish call

stmcginnis:
Ah, you're right. Looks like we had to do that to make HPE happy.

So I think this PR is part of the way there. If supported types is empty, then we could check if there is an action info link and make an extra call to query the system for supported types. Or to minimize overhead, I guess the presence of that action info link could be enough to differentiate from the HPE systems that don't report anything at all, so if so just put the `ResetType` in the payload.

Actually querying the system using the link would be more correct, just thinking about minimizing that round trip time. If you want to try things out and see what works best, I could go either way.

Thanks!

iamsli:
I agree with you on minimizing roundtrip time and implemented it so that it just checks if actionInfo is there. LMK what you think!

stmcginnis (approved):
Thanks!

PR #348, opened 2024-06-11
Add OEM field for ComputerSystem

lukassup:
I'd like to add `ComputerSystem.OEM` field as HPE and Dell provide it.

It's not part of the Redfish API schema so not sure if this is the right way to do it.

### DIFF
--- a/redfish/computersystem.go
+++ b/redfish/computersystem.go
@@ -825,6 +825,9 @@ type ComputerSystem struct {
 	// networkInterfaces shall be a link to a collection of type
 	// NetworkInterfaceCollection.
 	networkInterfaces string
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks, this looks good. You then should be able to follow[ the pattern used for the event service](https://github.com/stmcginnis/gofish/blob/main/oem/dell/eventservice.go#L58) to transform a standard `ComputerSystem` to an OEM one. That can happen either here in the `oem/` folder, or you can define your own local structs. Or just directly access the data out of the OEM field, of course.

lukassup:
Thanks, that was quick!

I can see that in this repo the structs are generated from Redfish schema. I'm thinking whether the newly generated structs won't clear this field and how to preserve it from being overridden?

PR #350, opened 2024-07-01
r/hostinterface.go: Correct "CredentialBootstrapping" type

arossbell:
Attempted use of Hostinterface (ie `redfish.GetHostInterface(c, uri)`) led to error:
```
json: cannot unmarshal object into Go struct field .CredentialBootstrapping of type string
```

The CredentialBootstrapping variable may have intended to be of type "CredentialBootstrapping" (struct defined earlier), which was not referenced elsewhere in gofish.

This commit resolves the unmarshal issue.


stmcginnis (approved):
Thanks! I did check the schema and you are correct, this should be the object. 🎉

Issue #349, opened 2024-06-28
http requests to keep using the same TCP connection


valdemarpavesi:
using context and setting IdleConnTimeout and DisableKeepAlives and req.Close = false

still release the TCP for every new request.

We wanna keep the same TCP connection
Have you accomplished it?

```
        ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
	defer cancel() // Ensure the context is canceled to free resources

	defaultTransport := http.DefaultTransport.(*http.Transport)
	transport := &http.Transport{
		Proxy: defaultTransport.Proxy,
		//DialContext:           defaultTransport.DialContext,
		DialContext: (&net.Dialer{
			Timeout:   60 * time.Second,
			KeepAlive: 15 * time.Second,
			DualStack: true,
		}).DialContext,
		MaxIdleConns:          defaultTransport.MaxIdleConns,
		IdleConnTimeout:       10 * time.Minute, //defaultTransport.IdleConnTimeout
		ExpectContinueTimeout: defaultTransport.ExpectContinueTimeout,
		TLSHandshakeTimeout:   30 * time.Second,
		DisableKeepAlives:     false, // default true, if true, disables HTTP keep-alives and will only use the connection to the server for a single HTTP request, tcp rst for every send.
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: true,
		},
	}

	client_HTTPClient := &http.Client{Transport: transport}

	config := gofish.ClientConfig{
		Endpoint:   "test",
		Username:   "test",
		Password:   "test",
		Insecure:   true,
		HTTPClient: client_HTTPClient,
	}

	c, err := gofish.ConnectContext(ctx, config)
	if err != nil {
		panic(err)
	}

	service := c.Service
	chassis, err := service.Chassis()
	if err != nil {
		panic(err)
	}

	for _, chass := range chassis {
		fmt.Printf("Chassis: %#v\n\n", chass)
	}

	// The client and its underlying connections can be reused for subsequent requests
	// The connections will be kept open and reused according to the transport's configuration
	fmt.Println("Press any key to continue...")
	var input string
	fmt.Scanln(&input)
```

stmcginnis:
I have not tried this, but you could make a local modification and see if it would work. I _think_ you would just need to change this line:

https://github.com/stmcginnis/gofish/blob/main/client.go#L503

valdemarpavesi:
Thanks @stmcginnis  

[...]

PR #351, opened 2024-07-02
Expose config option to keep alive TCP connections

stmcginnis:
This adds a `ReuseConnections` configuration flag. When set, this will try to configure the HTTPClient transport to not immediately close connections so that they can be reused by subsequent API calls.

Closes: #349 

### Testing

I ran something that collects a lot of different information from a system. Its original unmodified timing was:

```txt
real	0m35.706s
user	0m0.165s
sys	0m0.063s
```

After setting the new reuse flag, the same operations timing was:

```txt
real	0m26.284s
user	0m0.266s
sys	0m0.083s
```

I also tried tweaking both `ReuseConnections: true` and `MaxConcurrentRequests: 2` and got even better results:

```txt
real	0m21.510s
user	0m0.092s
sys	0m0.039s
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -52,6 +52,9 @@ type APIClient struct {
 
 	// dumpWriter will receive HTTP dumps if non-nil.
 	dumpWriter io.Writer
+
+	// keepAlive is a flag to indicate if we should try to keep idle connections open
+	keepAlive bool
 }
 
 // Session holds the session ID and auth token needed to identify an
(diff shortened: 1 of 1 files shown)


stmcginnis:
Local testing is good, so I'm going to merge this. But I would still appreciate any feedback from others to make sure this seems to work as expected and to know if there are any issues with this approach that I've overlooked. Please open an issue if there are any negative side effects from this that I've overlooked.

PR #352, opened 2024-07-04
Use common generic fetch of collection objects

stmcginnis:
This moves the minimum supported Go version to 1.18 to allow the use of gneerics in the code.

This adds a common function for getting all objects in a collection referenced by a link. This is used by most objects to retrieve a set of all objects of the given type. With the ability to use generics now, this eliminates a lot of repetitive code.

### DIFF
--- a/common/collection.go
+++ b/common/collection.go
@@ -138,3 +138,47 @@ func CollectCollection(get func(string), links []string) {
 
 	wg.Wait()
 }
+
+func GetCollectionObjects[E any](c Client, link string, getFn func(Client, string) (*E, error)) ([]*E, error) {
+	var result []*E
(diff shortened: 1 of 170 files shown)


common/collection.go
```diff
@@ -138,3 +138,47 @@ func CollectCollection(get func(string), links []string) {
...
+
+	for r := range ch {
+		if r.Error != nil {
+			collectionError.Failures[r.Link] = r.Error
+		} else {
+			result = append(result, r.Item)
+		}
+	}
+
+	fmt.Printf("Got here, %d objects in collection\n", len(result))
```
stmcginnis, line 177: Oops, will fix shortly with some other changes.

PR #354, opened 2024-07-08
Add common concurrent function to get list of objects

stmcginnis:
We have concurrency when getting a collection that contains a set of links to objects. But some object properties are a list themselves and not a Collection object. This adds a common function to be used for these types of properties to take advantage of the concurrent processing to fetch these objects from the system.


stmcginnis:
As an accedotal data point, I have something that collects a lot of different information from a system. Prior to this change it usually took 20-30 seconds to complete. After these changes it appears to take a little under 9 seconds.

PR #356, opened 2024-07-11
Add ability to get the RedundancySet for a given Redundancy group in the Power resource

pereztr5:
This change makes it so we can access the `RedundancySet` param, which contains all actual Power Supplies vs what is listed in `.PowerSupplies` as that can contain anything in the PSU slots, which sometimes are fans and not actual power supplies.

### DIFF
--- a/redfish/power.go
+++ b/redfish/power.go
@@ -179,6 +179,24 @@ func (power *Power) PowerSupplyReset(memberID string, resetType ResetType) error
 	return power.Post(power.powerSupplyResetTarget, t)
 }
 
+// RedundancySet returns the []PowerSupply in the target redundancy group based on the
+// memberID. The memberID is the ID for the redundancy group from the list of .Redundancy
+// from /redfish/v1/Chassis/{id}/Power
+func (power *Power) RedundancySet(memberID int) []PowerSupply {
+	var powerSupplies []PowerSupply
+	if len(power.Redundancy) >= memberID+1 {
+		for _, psLink := range power.Redundancy[memberID].redundancySet {
+			for i := range power.PowerSupplies {
+				if power.PowerSupplies[i].ODataID == psLink {
+					powerSupplies = append(powerSupplies, power.PowerSupplies[i])
+				}
+			}
+		}
+	}
+
+	return powerSupplies
+}
+
 // GetPower will get a Power instance from the service.
 func GetPower(c common.Client, uri string) (*Power, error) {
 	return common.GetObject[Power](c, uri)


stmcginnis (changes requested):
Please expose this by adding an accessor method so it is consistent with the other object/related-object acccess.

For example:
https://github.com/stmcginnis/gofish/blob/main/redfish/power.go#L425

Thanks for working on this!

redfish/power.go
```diff
@@ -179,6 +179,20 @@ func (power *Power) PowerSupplyReset(memberID string, resetType ResetType) error
 	return power.Post(power.powerSupplyResetTarget, t)
 }
 
+// GetRedundancySet return the []PowerSupply in the target redundancy group based on the memberID.
+func (power *Power) GetRedundancySet(memberID int) []PowerSupply {
```
stmcginnis, line 183: Please rename the method to just be `RedundancySet(memberID int)`.

redfish/power.go
```diff
@@ -179,6 +179,20 @@ func (power *Power) PowerSupplyReset(memberID string, resetType ResetType) error
...
+	var powerSupplies []PowerSupply
+	for _, psLink := range power.Redundancy[memberID].redundancySet {
```
stmcginnis, line 185: Would probably be good to make sure `memberID` is actually present before accessible `.redundancySet`.

stmcginnis (commented):
Thanks - just a couple minor things.

pereztr5:
Hey @stmcginnis, I'm unsure if I made this function consistent with your other object/related-object access as I could not use the `redfish.Redundancy[].redundancySet` links. These links look like:
```
# /redfish/v1/Chassis/1/Power | jq '.Redundancy[].RedundancySet'
[
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/0"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/1"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/2"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/3"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/4"
  },
  {
    "@odata.id": "/redfish/v1/Chassis/1/Power#/PowerSupplies/7"
  }
]
```

Hitting those links gives back the same response as `/redfish/v1/Chassis/1/Power` vs. the individual `redfish.PowerSupply`, and so in both cases, you need to use the `.PowerSupplies[]` field from the response to get the actual `redfish.PowerSupply`.

My goal is to get the `[]redfish.PowerSupply` for the given redundancy group so that it does not include anything that is not part of it. That is why I made the function just filter the `[]redfish.PowerSupply`.

stmcginnis:
> My goal is to get the []redfish.PowerSupply for the given redundancy group so that it does not include anything that is not part of it. That is why I made the function just filter the []redfish.PowerSupply.

Yep, that makes sense. Those are jsonpath links, so in this case they are self-referential. I think the approach makes sense.

pereztr5:
Thanks for the quick feedback. I have made the adjustments and added more to the comment.

stmcginnis (approved):
Thanks for the updates - looks good!

PR #360, opened 2024-09-02
Change Memory Struct (NonVolatileSizeLimitMiB)

DeathwingTheBoss:
Since Redfish DMTF Data Spec 2022.2 the Memory field `NonVolatileSizeLimitMiB` is supposed to be type `integer` but Gofish has it as `string` resulting in unmarshalling errors.

This fixes that.

(Had a problem locally, wanted to help upstream in case someone runs into the same issue eventually)


stmcginnis (approved):
Thanks!

PR #366, opened 2024-10-09
redfish/drive: expose oem data and actions

Matt1360:
We have some nodes that don't follow the typical indicator light flow that `Update()` would handle. Instead, we have to hit a linked reference (with a variable key, sigh..) from the OEM section.

I'm hoping to expose the `Oem` field in both the `Drive` and the `Drive.Actions` (while also exposing the `Actions`) so that the user can follow that chain when necessary. This allows us to get the right link to where we need to send the `POST` or `PATCH` (also variable, heh) to enable lights for our datacenter folks.

Tests updated to not lose secure erase functionality.

### DIFF
--- a/oem/smc/drive.go
+++ b/oem/smc/drive.go
@@ -1,0 +1,62 @@
+package smc
+
+import (
+	"encoding/json"
+
+	"github.com/stmcginnis/gofish/redfish"
+)
+
+type Drive struct {
+	redfish.Drive
+	Oem DriveOem `json:"Oem"`
+}
+
+type DriveOem struct {
+	Supermicro struct {
+		Temperature             int
+		PercentageDriveLifeUsed int
+		DriveFunctional         bool
+	} `json:"Supermicro"`
+}
+
+type DriveTarget struct {
+	Target     string `json:"target"`
+	ActionInfo string `json:"@Redfish.ActionInfo"`
+}
+
+type DriveActions struct {
+	redfish.DriveActions
+	Oem struct {
+		DriveIndicate    DriveTarget `json:"#Drive.Indicate"`
+		SmcDriveIndicate DriveTarget `json:"#SmcDrive.Indicate"`
+	} `json:"Oem"`
+}
+
+func FromDrive(drive *redfish.Drive) (Drive, error) {
+	var oem DriveOem
+	err := json.Unmarshal(drive.Oem, &oem)
+
+	return Drive{
+		Drive: *drive,
+		Oem:   oem,
+	}, err
+}
+
+func FromDriveActions(da *redfish.DriveActions) (DriveActions, error) {
+	oemActions := DriveActions{
+		DriveActions: *da,
+	}
+
+	err := json.Unmarshal(da.Oem, &oemActions.Oem)
+	return oemActions, err
+}
+
+// DriveIndicateTarget checks both the SmcDriveIndicate and DriveIndicateTarget
+// Oem entries and returns the first populated target, due to key inconsistencies
+func (da DriveActions) DriveIndicateTarget() string {
+	if len(da.Oem.SmcDriveIndicate.Target) > 0 {
+		return da.Oem.SmcDriveIndicate.Target
+	}
+
+	return da.Oem.DriveIndicate.Target
+}
--- a/oem/smc/drive.go
+++ b/oem/smc/drive.go
@@ -2,38 +2,28 @@ package smc
 
 import (
 	"encoding/json"
+	"errors"
 
+	"github.com/stmcginnis/gofish/common"
 	"github.com/stmcginnis/gofish/redfish"
 )
+
+// ErrActionNotSupported is returned when the requested OEM-specific action
+// does not appear to be supported. This might happen when a device is new
+// or upgraded to a new firmware that follows the DMTF standards.
+var ErrActionNotSupported = errors.New("oem-specific action unsupported")
 
 // Drive extends a redfish.Drive for additional OEM fields
 type Drive struct {
 	redfish.Drive
-	smc struct {
-		Oem     driveOem     `json:"Oem"`
-		Actions driveActions `json:"Actions"`
-	}
-}
 
-type driveOem struct {
-	Supermicro struct {
-		Temperature             int
-		PercentageDriveLifeUsed int
-		DriveFunctional         bool
-	} `json:"Supermicro"`
-}
+	// Fields from the SMC OEM section
+	temperature             int
+	percentageDriveLifeUsed int
+	driveFunctional         bool
 
-type driveTarget struct {
-	Target     string `json:"target"`
-	ActionInfo string `json:"@Redfish.ActionInfo"`
-}
-
-type driveActions struct {
-	redfish.DriveActions
-	Oem struct {
-		DriveIndicate    driveTarget `json:"#Drive.Indicate"`
-		SmcDriveIndicate driveTarget `json:"#SmcDrive.Indicate"`
-	} `json:"Oem"`
+	// indicateTarget is the uri to hit to change the light state
+	indicateTarget string
 }
 
 // FromDrive returns an OEM-extended redfish drive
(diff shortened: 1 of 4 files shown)


Matt1360:
:+1: I'll take a look at adding it over there. For SMC, should I follow the other pending change and use `oem/supermicro`, or should it be abbreviated as `oem/smc`?

stmcginnis:
Hmm, good question. I don't have a strong opinion either way really. I think most folks who need it won't have any confusion equating smc to being for SuperMicro, so I'm fine with whatever you feel like doing.

Thanks!

Matt1360:
I've added a commit that introduces the `oem/smc` package, making use of the OEM fields exposed in the first commit, along with a simple test. Is this what you had in mind? I kept the first commit in order to still expose the OEM fields in both the `Drive` and the `DriveActions` structs.

oem/smc/drive.go
```diff
@@ -0,0 +1,62 @@
...
+		DriveActions: *da,
+	}
+
+	err := json.Unmarshal(da.Oem, &oemActions.Oem)
+	return oemActions, err
+}
+
+// DriveIndicateTarget checks both the SmcDriveIndicate and DriveIndicateTarget
+// Oem entries and returns the first populated target, due to key inconsistencies
+func (da DriveActions) DriveIndicateTarget() string {
```
stmcginnis, line 56: Move this to be a function on `Drive` directly and get rid of `DriveActions`.

oem/smc/drive.go
```diff
@@ -0,0 +1,62 @@
...
+	var oem DriveOem
+	err := json.Unmarshal(drive.Oem, &oem)
+
+	return Drive{
+		Drive: *drive,
+		Oem:   oem,
+	}, err
+}
+
+func FromDriveActions(da *redfish.DriveActions) (DriveActions, error) {
```
stmcginnis, line 45: Get rid of this and just have this part of the `Drive` unmarshaling in `FromDrive`.

oem/smc/drive.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/stmcginnis/gofish/redfish"
+)
+
+type Drive struct {
+	redfish.Drive
+	Oem DriveOem `json:"Oem"`
+}
+
+type DriveOem struct {
```
stmcginnis, line 14: Don't have these three be public structs. They can be used for unmarhsalling the `Drive` data, but they shouldn't be exposed beyond that.

redfish/drive.go
```diff
@@ -299,6 +299,10 @@ type Drive struct {
 	// WriteCacheEnabled shall indicate whether the drive
 	// write cache is enabled.
 	WriteCacheEnabled bool
+	// Actions are all the listed actions under the drive
+	Actions DriveActions
```
stmcginnis, line 303: Remove this, actions shouldn't be exposed directly.

stmcginnis (commented):
Great start. A few issue with consistency with how we do things in Gofish, so just a few minor changes needed.

Thanks!

redfish/drive.go
Matt1360: With this removed, how should I surface the `Oem` field of the actions to the user? I could expose `ActionsOem` as a `json.RawMessage` off the `Drive`, perhaps? I'll need that `Oem` field under actions from the drive somehow exposed so the `oem/smc` package can consume it.

oem/smc/drive.go
Matt1360: I have this staged and expose the attributes within via getters to keep them from being exposed.

```go
func (d Drive) Temperature() int {
        return d.smc.Oem.Supermicro.Temperature
}
```

oem/smc/drive.go
stmcginnis: Also, this should not be returning the URL. That URL should just be used internally as the target to use for calling this action.

redfish/drive.go
stmcginnis: It shouldn't be surfaced to the end user. Just in the oem object, there should be functions on the OEM Drive to perform the available actions. So something like:

```go
var drive redfish.Drive // Assigned elsewhere

smcDrive, err := smc.FromDrive()
smcDrive.Indicate()
```

Matt1360: Understood; let the OEM Drive model perform the action. The confusion I'm having here is that the URL that's necessary is buried in the Actions, under another `Oem` field as seen in the test [here](https://github.com/stmcginnis/gofish/pull/366/files#diff-d2586b9d6c105ec199b20cfd0957883d3d6be91e06be8171843823a21f21914eR38-R45).

The `oem/smc` package needs a way to get the drive -> actions -> oem data. That's why I exposed the `Actions` struct here, or suggest exposing `ActionsOem` hanging off the `redfish.Drive`. I'm not sure what approach best fits the gofish conventions.

stmcginnis: Yeah, sorry, I didn't have a good example to easily show what I was thinking. I've done a couple SMC updates, so now I do have good examples.

Here's an example of accessing the properties of an OEM object: https://github.com/stmcginnis/gofish/blob/main/oem/smc/chassis.go#L27-L44

And its usage: https://github.com/stmcginnis/gofish/blob/main/oem/smc/chassis_test.go#L105

I don't have a simple example of using an action target, but it would be similar to having the function on the OEM struct like here: https://github.com/stmcginnis/gofish/blob/main/oem/smc/nodemanager.go#L100

And pulling out the action target like you would a property or link similar to the first example or here: https://github.com/stmcginnis/gofish/blob/main/oem/smc/computersystem.go#L51

Hopefully that's a little clearer. There are other examples in the standard redfish objects that show some similar approaches. Basically, I'd like to keep these OEM objects so they have a similar feel and behavior to the regular ones. I'm seeing now that some of the earlier OEM objects added did not do that, so sorry if you were using those an example. I should probably fix that. ;)

Matt1360: Thanks! I wasn't finding a single answer, but a few ways to do it as you said.

In the `Chassis` one, it seems that the `RawData` is exported in the `redfish.Chassis`, which isn't the case in the `redfish.Drive`. If I change that to be exported that will make this follow that path much easier. I'll push what I have now so you can see the state it's in, and I'll clean it up a bit for the next pass.

stmcginnis: Perfect, thanks! I don't love having RawData exposed, but it seems like a decent trade-off to make these work easier.

Matt1360: Pushed this up. I previously added the helper functions for getting the temperature etc, but I could remove those getters and simply export the values if you think that's more appropriate here.

oem/smc/drive.go
```diff
@@ -0,0 +1,94 @@
...
+	redfish.Drive
+
+	// Fields from the SMC OEM section
+	temperature             int
+	percentageDriveLifeUsed int
+	driveFunctional         bool
+
+	// indicateTarget is the uri to hit to change the light state
+	indicateTarget string
+}
```
stmcginnis, line 27:
```suggestion
type Drive struct {
	redfish.Drive

	// Fields from the SMC OEM section
	Temperature             int
	PercentageDriveLifeUsed int
	DriveFunctional         bool

	// indicateTarget is the uri to hit to change the light state
	indicateTarget string
}
```

oem/smc/drive.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// Populate the Oem data
+	if err := json.Unmarshal(drive.RawData, &t); err != nil {
+		return smcDrive, err
+	}
+
+	smcDrive.temperature = t.Oem.Supermicro.Temperature
+	smcDrive.percentageDriveLifeUsed = t.Oem.Supermicro.PercentageDriveLifeUsed
+	smcDrive.driveFunctional = t.Oem.Supermicro.DriveFunctional
```
stmcginnis, line 58:
```suggestion
	smcDrive.Temperature = t.Oem.Supermicro.Temperature
	smcDrive.PercentageDriveLifeUsed = t.Oem.Supermicro.PercentageDriveLifeUsed
	smcDrive.DriveFunctional = t.Oem.Supermicro.DriveFunctional
```

oem/smc/drive.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// We check both the SmcDriveIndicate and the DriveIndicate targets
+	// in the Oem sections - certain models and bmc firmwares will mix
+	// these up, so we check both
+	smcDrive.indicateTarget = t.Actions.Oem.DriveIndicate.Target
```
stmcginnis, line 63: 🎉

oem/smc/drive.go
```diff
@@ -0,0 +1,94 @@
...
+
+// PercentageDriveLifeUsed returns the OEM provided drive life estimate as a percentage used
+func (d Drive) PercentageDriveLifeUsed() int {
+	return d.percentageDriveLifeUsed
+}
+
+// Functional returns the OEM provided flag that suggests whether a drive is functional or not
+func (d Drive) Functional() bool {
+	return d.driveFunctional
+}
```
stmcginnis, line 84:
```suggestion

```

stmcginnis (commented):
Really looking great. Love the way you are handling the changing action name so the library user doesn't need to be concerned about it.

Only thing I think is to just expose the new properties directly rather than getting them through an accessor.

redfish/drive.go
stmcginnis: Just saw this comment before posting. Yes, I think that would be great.

oem/smc/drive.go
```diff
@@ -0,0 +1,79 @@
+package smc
```
stmcginnis, line 1: Oops, need to include the copyright header:

```go
//
// SPDX-License-Identifier: BSD-3-Clause
//

package smc
```

Matt1360: Thanks!

stmcginnis:
Nice work - thanks a lot!

Matt1360:
No problem! Do you have a sense for when the next release will be cut? Deciding whether to use my fork for now, or wait a bit and bump the version.

stmcginnis:
You can use `go get github.com/stmcginnis/gofish@4b8c27ed6527ba5cd775d82feb2f58198a33e3a2` to use the current HEAD. That's what I usually recommend.

Otherwise, I'm probably going to add a few more Supermicro OEM things since I now have access to the hardware, then tag a release once I get at least some of the main objects in place. So I guess whatever is more convenient for you.

PR #379, opened 2024-10-22
Increase Sensor.Reading to float64

iamsli:
We are seeing some implementations using float64 for the generic Sensor.Reading property


stmcginnis (approved):
Thanks!

PR #381, opened 2024-10-23
Gets instances of resources where relevant in cooling unit

iamsli:
fixes 
- `Reservoirs` typo causes bad json unmarshal
- `EnvironmentMetrics` and `LeakDetections` are not collections but resources of cooling unit


stmcginnis (approved):
🎉

PR #382, opened 2024-11-11
Expose ManagersProvidingService link from service root

iamsli:
I'd like to be able to just see the uri of the managers link from service root without calling the url.

### DIFF
--- a/serviceroot.go
+++ b/serviceroot.go
@@ -192,9 +192,9 @@ type Service struct {
 
 	// Sessions shall contain the link to a collection of Sessions.
 	sessions string
-	// ManagerProvidingService shall contain a link to a resource of type Manager that represents the manager providing
+	// ManagerProvidingServiceLink shall contain a link to a resource of type Manager that represents the manager providing
 	// this Redfish service.
-	managerProvidingService string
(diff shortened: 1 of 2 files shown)


stmcginnis:
We've generally tried not to expose the links directly throughout gofish. Can you explain why you would need something like this in the library? Especially from the service root, this seems easy enough to get with a couple of http get requests.

iamsli:
I want to be able to know what the device is from the service root independent of other properties. Definitely can get this via some http gets but wanted to make the code simpler 😄

iamsli:
@stmcginnis feel free to close this PR if it goes against the integrity of gofish

stmcginnis:
Sorry, I know it's a pretty trivial thing, but I'd like to try to avoid this unless there's a strong need. 

I think some vendors have OEM extensions to the ServiceRoot. It's a little more clunky, but I could see we might need to have a `RawData` property like some of the other objects so an OEM version of the ServiceRoot could be instantiated. Maybe you could add that and pull out the link from there?

iamsli:
sounds like this isn't a good fit for gofish, closing!

PR #387, opened 2024-12-02
Add Conditions to Status

### LINKED ISSUE #384: Status is missing Condition
As of v1.11 there is a Condition property as part of Status. This is missing from the library today and we are starting to see this implemented.

iamsli:
as of DSP2046 2024.3 Status contains a Conditions property

fixes #384


stmcginnis (approved):
Thanks @iamsli!

PR #389, opened 2024-12-05
Add method to get metrics from PowerSupply

smiller248:
It appears that there is both a [PowerSupply](https://pkg.go.dev/github.com/stmcginnis/gofish/redfish#PowerSupply) and a [PowerSupplyUnit](https://pkg.go.dev/github.com/stmcginnis/gofish/redfish#PowerSupplyUnit) type, which both represent the same Redfish "PowerSupply" resource.

Unfortunately, `PowerSupply` and `PowerSuppyUnit` each have some features that the other does not. For my current use case, I need some fields that only exist on `PowerSupply` but also require the `Metrics()` method, which is only available on `PowerSupplyUnit`.

I have opted for that path of least resistance here: just adding a link to `PowerSupplyUnitMetrics` from `PowerSupply` (since it's the same underlying Redfish resource schemas, everything parses correctly). Note: there is no `PowerSupplyMetrics`.

Please let me know if:
1. There is a good reason that `PowerSupply` and `PowerSupplyUnit` are distinct, which am oblivious of
2. You would prefer a more comprehensive fix (probably with breaking changes) rather than the path-of-least-resistance option

### DIFF
--- a/redfish/power.go
+++ b/redfish/power.go
@@ -367,6 +367,8 @@ type PowerSupply struct {
 	// member within the collection. For services supporting Redfish v1.6 or
 	// higher, this value shall be the zero-based array index.
 	MemberID string `json:"MemberId"`
+	// The link to the power supply metrics resource associated with this power supply.
+	metrics string
 	// Model shall contain the model information as defined
 	// by the manufacturer for the associated power supply.
 	Model string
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Looks like I didn't review closely enough when `PowerSupplyUnit` was added. That is indeed just a `PowerSupply`.

We _should_ just collapse its content into `PowerSupply` and get rid of it. That may break some folks, but it makes sense to do. And since Gofish is still a `v0.x` release, I don't feel too badly about it.

This is good for now. If you'd like to follow up with more, that would be awesome. Otherwise I'll put it on my list to try to get to some day.

Thanks!

PR #391, opened 2024-12-12
Ports gets the instances managed by the Processor

iamsli:
per the spec Ports is defined as a link to the collection of the resource. In keeping with the pattern throughout the Gofish lib Ports() should return the instances of the Ports under processors. (Mostly I've copied the pattern established with `client.serviceroot.Chassis()` :)


stmcginnis (approved):
Looks great, thanks!

PR #396, opened 2025-01-16
Clean up session on logout

### LINKED ISSUE #395: Session Logout must remove session ID and token in `auth` struct?
Hello,

When we perform [`Logout`](https://github.com/stmcginnis/gofish/blob/47866ef9f6c79a764843f67f9af336311a27d9c5/client.go#L594-L599) to a session, the `c.auth.Session` and `c.auth.Token` must be also removed, right?! So that we can use [`CloneWithSession`](https://github.com/stmcginnis/gofish/blob/47866ef9f6c79a764843f67f9af336311a27d9c5/client.go#L255) to establish a new session. 

In long running services like monitoring, if sessions get invalidated for any reason, we can simply do a Logout and re-establish a new session using `CloneWithSession` method by implementing the proposed change.

Does it make sense? If not, is there any better way to re-establish a session when existing session gets invalidated?

Cheers!

mahendrapaipuri:
Closes #395

### DIFF
--- a/client.go
+++ b/client.go
@@ -593,7 +593,12 @@ func (c *APIClient) dumpResponse(resp *http.Response) error {
 // a new connection.
 func (c *APIClient) Logout() {
 	if c != nil && c.Service != nil && c.auth != nil {
-		_ = c.Service.DeleteSession(c.auth.Session)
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks!

PR #398, opened 2025-01-17
Fix missing ResetType value

pereztr5:
In my previous PR(#394) I forgot to actually add the reset type value of `FullPowerCycle`.


stmcginnis (approved):
Oops! :)

PR #400, opened 2025-01-21
add MediaType

Muyk33rus:


Muyk33rus:
@stmcginnis and this PR)

redfish/virtualmedia.go
```diff
@@ -134,6 +134,8 @@ type VirtualMedia struct {
 	ImageName string
 	// Inserted shall indicate whether media is present in the virtual media device.
 	Inserted bool
+	// Сonnection MediaType used
+	MediaType VirtualMediaType
```
stmcginnis, line 138: This isn't a [property of the VirtualMedia object](https://redfish.dmtf.org/schemas/v1/VirtualMedia.v1_6_4.yaml). Can you explain why this is needed here?

Muyk33rus: Yes of course, look, I'll give you an example from a server where this is:
{
    "@odata.id": "/redfish/v1/Managers/bmc/VirtualMedia/USB1",
    "@odata.type": "#VirtualMedia.v1_3_0.VirtualMedia",
    "Actions": {
        "#VirtualMedia.EjectMedia": {
            "target": "/redfish/v1/Managers/bmc/VirtualMedia/USB1/Actions/VirtualMedia.EjectMedia"
        },
        "#VirtualMedia.InsertMedia": {
            "target": "/redfish/v1/Managers/bmc/VirtualMedia/USB1/Actions/VirtualMedia.InsertMedia"
        }
    },
    "Id": "USB1",
    "MediaType": "USBStick",
    "MediaTypes": [
        "CD",
        "USBStick",
        "HDD"
    ],
    "Name": "Virtual Removable Media",
    "Oem": {
        "OpenBMC": {
            "@odata.type": "#OemVirtualMedia.v1_0_0.VirtualMedia"
        }
    },
    "TransferMethod": "Stream",
    "WriteProtected": true
}
in https://redfish.dmtf.org/schemas/v1/VirtualMedia.v1_3_0.json

Muyk33rus: @stmcginnis  Ping (=

stmcginnis: Hmm, that property didn't exist in the [1.3.0 version of the object](https://redfish.dmtf.org/schemas/v1/VirtualMedia.v1_3_0.yaml) either (which, by the way, is quite old now).

I looked back to the 1.0.0 version of the VirtualMedia object to see if this was a deprecated field that was removed, but it doesn't exist there either.

It might be better to upgrade your BMC firmware, and/or file an issue with your vendor to let them know that they are not complying to the spec.

Muyk33rus: so in 1.6.4 this field is also present 
Снимок экрана 2025-01-27 в 17 41 46

stmcginnis: That is the type definition. You need to look under `VirtualMedia.properties`.

Muyk33rus: Снимок экрана 2025-01-27 в 17 45 29

Muyk33rus: https://github.com/DMTF/Redfish-Publications/blob/main/json-schema/VirtualMedia.v1_6_4.json#L180 look this please

Muyk33rus: Doesn't it bother you that a string type is specified?

stmcginnis: https://github.com/DMTF/Redfish-Publications/blob/main/json-schema/VirtualMedia.v1_6_4.json#L180 points to the type definition. It is then used as the data type of the [VirtualMedia.MediaTypes](https://github.com/DMTF/Redfish-Publications/blob/main/json-schema/VirtualMedia.v1_6_4.json#L387) property. You can see under `VirtualMedia.properties` definitions there is only the one plural `MediaTypes` property defined using the `MediaType` data type.

I'm not sure if I understand what you are referring to with having a string type. Can you expand on that? Thanks!

Muyk33rus: If the device is not connected, all possible connection types are displayed. However, when the device is connected, it shows the specific type through which it is connected. This is so that you can later launch this virtual disk through One Time Boot or set the appropriate priority in the BIOS settings for the drive. If this parameter were not present, you would not know which type is being used for the connection.

stmcginnis:
This was picked up with other changes.

PR #402, opened 2025-02-04
Add client to the returned ManagerAccount on account creation

pereztr5:
The returned `*ManagerAccount` when using `CreateAccount()` did not have a client set which made using `*ManagerAccount` have a nil reference when attempting to make HTTP calls afterwards.


stmcginnis:
> This works for most common BMCs but for some OpenBMC versions it seems to want the If-Match or If-None-Match set. I am not sure the best way to get those headers set.

Thanks! I think that's a larger issue (of many) with OpenBMC. This is a good fix for now!

stmcginnis (approved):

PR #403, opened 2025-02-12
LeakDetectors gets the collection in chassis

iamsli:
whoops now realizing that I made a mistake. LeakDetectors is a collection under Chassis 😄


stmcginnis (approved):
At least I'm not the only one. 😂

PR #406, opened 2025-02-25
Add Oem property to MemoryMetrics

mwieczorek:
This PR adds `Oem` property for `MemoryMetrics` resource, similarly like in `ProcessorMetrics`.


mwieczorek:
lint job failures should be fixed in https://github.com/stmcginnis/gofish/pull/407

stmcginnis (approved):

PR #407, opened 2025-02-25
Fix golangci config

mwieczorek:
Lint job started to fail complaining about jsonschema of config (f.e. one of my previous [PRs is ok](https://github.com/stmcginnis/gofish/actions/runs/13303597831/job/37149525323?pr=404#step:1:38), [the other fails](https://github.com/stmcginnis/gofish/actions/runs/13518590958/job/37772547791?pr=406#step:1:38))

Lint job uses v6 version of action and it was updated recently with this PR: https://github.com/golangci/golangci-lint-action/pull/1171 - so the job started to verify the config's schema.

This PR fixes the config:
- removes unsupported property for `nolintlint` - https://golangci-lint.run/usage/linters/#nolintlint
- changes from array to object in `goheader` - https://golangci-lint.run/usage/linters/#goheader

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -19,7 +19,7 @@ linters-settings:
   goheader:
     values:
       const:
-        - LICENSE: BSD-3-Clause
+        LICENSE: BSD-3-Clause
     template: |-
       SPDX-License-Identifier: {{ LICENSE }}
   goconst:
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks @mwieczorek!!

Issue #340, opened 2024-05-20
Use the $expand parameter when querying collections


Arno500:
When querying subcollections, gofish is currently iterating manually 3 by 3 to get the data. This is painfully slow, especially when the collections are big and the server is far away.

I would propose to include support for the $expand OData parameter when supported (this needs to be checked when querying a collection for the first time on this entity). It would allow the device to answer once for all the sub elements when asked, making the query much faster and reducing the impact of latency.

My specific case was requesting a PDU with 48 plugs, that is pretty slow in Redfish already. Querying all of them individually is not exactly fun as you can guess :P

stmcginnis:
Thanks for bringing this up. It is an interesting idea. The Redfish service can be a little slow responding, but that network latency of making multiple calls to get the elements is definitely the bigger bottleneck.

I haven't spent a ton of time on this, but what I had thought about doing was to set a flag if it's supported. Luckily, the `ServiceRoot` object that you get at the root of the object hierarchy should contain `ServiceRoot.ProtocolFeaturesSupported.ExpandQuery.ExpandAll`. So right away on connection it should be possible to determine if the service will support expanding collections or not.

The trickier part is probably including that in a uniform interface for the library consumer. But I think during the custom unmarshalling that we need to do, we could try to collect that array of sub-objects in a private property. Then the consumer would still call the same method as they would if expand wasn't being used (e.g. `outlets, err := pdu.Outlets()`). Then internally, if the expanded data is there we can just return the collection of objects. If it is not there, then we can use the existing method of getting the collection of objects the slower way.

I will try to get some time to try this out some time soon.

stmcginnis:
I honestly haven't had a lot of time to ponder this. Yes, PRs definitely welcome! I'm trying to remember now, but I was thinking we could have some flag set so the library user could opt in to operating in a way that could potentially consume a lot more memory. Then in the accessor methods (e.g. `computerSystem.VirtualMedia()`) it would determine if it already has the objects or if it needs to fetch them.

joeyberkovitz:
sounds good, I'd be happy to expand on #408 but @mwieczorek - LMK if you want to work on it

stmcginnis:
I think the main needs are covered by #482, so going to clean things up and close this. Feel free to reopen for anything missing.

PR #408, opened 2025-03-03
Add initial support for  in system.Memory()

mwieczorek:
This is a proposal of how to handle '$expand' redfish odata queries. Related issue: #340

Note: creating PR to get early feedback if general approach is valid, things to be added to this PR:

- [ ] add tests for queries with expand option
- [ ] godoc comments
- [ ] handle all Memory 'subresources', not only `Metrics`

Assumptions, I've made:
- I don't want to break existing clients of the pkg
- I don't want to change default behavior (in case someone update the pkg)

Because of above, I'd like to introduce the 'expand feature' with 'functional options' approach.
There's a new `common/odata.go` file with 2 'options' functions responsible for controlling the expand value and level (ref: [redfish spec](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.21.1.html#the-expand-query-parameter)).

Draft of this PR supports 'expand' only for `system.Memory()` and 'expands' only `Metrics`. It's possible to us it as it was (`system.Memory()`) or with expand option (`system.Memory(common.WithExpand(common.ExpandOptionPeriod))`). By default, no 'expand' is used, so we won't change the default behavior.
`Memory` struct was changed, and `metrics` var type is changed from `common.Link` to `MemoryMetrics`. In case the expand is not used for `system.Memory()`, `memory.Metrics()` call will make a call to redfish API.

Tested with Redfish mockup server and real device.

Please let me know if above approach is ok or if you have any suggestions how to do it differently.

---

Something I'm not sure how to tackle right now...
Following my use case (get system's memory and it's metrics):

```
system
  -> memory collection
    -> memory item
      -> metrics
```

`system.Memory()` encapsulate '2 levels' above: it send request to API to get 'memory collection' and then 'memory item'
Both of the requests can handle expand parameter. Let's assume Redfish API supports only 1 level for expand.
If we do `system.Memory(common.WithExpand(common.ExpandOptionPeriod))`, should gofish pass $expand param:
- only to 'memory collection'
- only to 'memory item'
- both?
Or should it be possible to decide when invoking `system.Memory()`

I assumed, when we use `system.Memory()` and get in response `[]Memory` we're interested in `Memory` resource and that's what expanded in this 'draft' PR. (tbh: that's what I'm interested in my use case - getting metrics for all 'memory items')
With this approach we don't change anything in `GetCollectionObjects` (except 'optional' list of `QueryOption` in the definition)

If we'd like to support that at 'memory collection' request, it will get more dificult and complex and would need more changes to those 'generic' functions (GetCollectionObjects -> CollectList -> GetCollection etc).

Let me know what you think and how you suggest to approach that.

### DIFF
--- a/common/collection.go
+++ b/common/collection.go
@@ -102,13 +102,13 @@ func (cr *CollectionError) Error() string {
 }
 
 // CollectList will retrieve a collection of entities from the Redfish service.
-func CollectList(get func(string), c Client, link string) error {
+func CollectList(get func(string), c Client, link string, queryOpts ...QueryOption) error {
 	collection, err := GetCollection(c, link)
 	if err != nil {
 		return err
 	}
 
-	CollectCollection(get, collection.ItemLinks)
+	CollectCollection(get, collection.ItemLinks, queryOpts...)
 	if collection.MembersNextLink != "" {
 		err := CollectList(get, c, collection.MembersNextLink)
 		if err != nil {
@@ -120,7 +120,7 @@ func CollectList(get func(string), c Client, link string) error {
 
 // CollectCollection will retrieve a collection of entitied from the Redfish service
 // when you already have the set of individual links in the collection.
-func CollectCollection(get func(string), links []string) {
+func CollectCollection(get func(string), links []string, queryOpts ...QueryOption) {
 	// Only allow three concurrent requests to avoid overwhelming the service
 	limiter := make(chan struct{}, 3)
 	var wg sync.WaitGroup
@@ -133,7 +133,7 @@ func CollectCollection(get func(string), links []string) {
 			defer wg.Done()
 			get(itemLink)
 			<-limiter
-		}(itemLink)
+		}(BuildQuery(itemLink, queryOpts...))
 	}
 
 	wg.Wait()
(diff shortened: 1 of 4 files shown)


stmcginnis:
Awesome, thanks for working on this. I'm probably tied up for a few days, but will try to dig into this as soon as I can.

mwieczorek:
I was thinking about $expand support in collections, and came to something like https://github.com/stmcginnis/gofish/pull/412

I created a separate PR in case it would be easier to compare both version for you.

In #412 it's possible to explicitly decide if we want to expand collection or objects in collection. It makes sense in case of only one level of expand is supported.

My use case:
I want to get 'memory metrics' and 'memory environment metrics' and reduce number of Redfish API calls.
Let's say we have 1 system, 1 memory 'item.
If we 'expand' collection (1 API request), we would still need to get 'metrics' and 'environmentMetrics' (so additional 2 API requests)
I we decide to expand only 'objects' (memory 'item'), then we will have 1 API request for collection and 1 API request for memory item.

Anyway, please take a look when you find time.

PS:
#412 makes changes in 'public' collections types, unfortunately I didn't find a way how to do it without such changes. The only thing is I guess those structs/funcs are used only 'internally' in the gofish so hopefully it wouldn't break someone's code.

stmcginnis:
Hey! Can you address the lint job errors?

Issue #409, opened 2025-03-04
panic: json: cannot unmarshal string into Go struct field .temp.Voltages.temp.UpperThresholdNonCritical of type float32


webfrank:
Error trying to extract Power information from CISCO CIMC

stmcginnis:
This is usually caused by the vendor implementation not following the spec and doing something like setting a numeric value to a string. Can you get the raw json response for the object so we can see what it's choking on?

stmcginnis:
Just to capture it here, this is the definition for that property:

```yaml
        UpperThresholdNonCritical:
          description: The value at which the reading is above normal range.
          nullable: true
          readOnly: true
          type: number
          x-longDescription: This property shall contain the value at which the `ReadingVolts`
            property is above the normal range.  The value of the property shall use
            the same units as the `ReadingVolts` property.
          x-units: V
```

So if they are returning it as a string, this will have to be raised with Cisco to fix their implementation.

webfrank:
Hi, yes, from raw I captured "N/A" as the value. I don't think Cisco will
soon fix it.
I understand the lib adheres to standard but should be able to handle this
kind of issues just ignoring them.
For now I'm going the http direct way but I would like to move on the lib
once a solution has been found.

Il mer 5 mar 2025, 14:31 Sean McGinnis ***@***.***> ha
scritto:

stmcginnis:
It can be handled, but I'm afraid there are a few similar fields here that would all need to be converted. Can you paste the formatted json that is returned from this system so I can see what they are doing?

webfrank:
Hi, this is the raw data:

```json
{
  "Id": "Power",
  "Name": "Power",
  "@odata.context": "/redfish/v1/$metadata#Chassis/Members/$entity/Power",
  "PowerControl": {
    "PhysicalContext": "PowerSupply",
    "PowerLimit": {
      "LimitException": "NoAction"
    },
    "PowerConsumedWatts": "152",
    "PowerMetric": {
      "IntervalInMin": 0.0833,
      "MinConsumedWatts": "113",
      "AverageConsumedWatts": "170",
      "MaxConsumedWatts": "240"
    }
  },
  "Voltages": [
    {
      "@odata.id": "/redfish/v1/Chassis/1/Power#/Voltages/PSU1_VOUT",
      "Status": {
        "State": "Enabled",
        "Health": "OK"
      },
      "SensorNumber": 41,
      "Name": "PSU1_VOUT",
[...]

PR #414, opened 2025-03-20
Workaround non-standard Cisco Power results

stmcginnis:
The JSON returned for these objects from certain Cisco systems does not conform to the spec. This attempts to work around the differences to bring it in line.

Closes: #409


stmcginnis:
@webfrank, can you confirm this works with your system?

PR #419, opened 2025-04-22
redfish/virtualmedia: Return Task if Link present in Insert response

### LINKED ISSUE #418: InsertMediaConfig task response is lost
I'm using the VirtualMedia.InsertMediaConfig function and recently added some extra fallback code that wasn't working out like I expected. I saw mention of `"Message":"InsertMedia/EjectMedia task is in Running state for other CD instance"`, which led me to learn that gofish is swallowing useful information from the response.

According to the spec [7.5.1 Data modification requests overview](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.0.html#modification-success-responses) we can expect a Location header in the response, I happen to see `Link` headers with the link to the task instead.

I'm not seeing a way to get at the task without having to search from the task service, which I'd like to avoid. I can't use `virtualmedia.PostWithResponse` because `insertMediaTarget` is private, so I think either make insertMediaTarget (and ejectMediaTarget) public or add a WithResponse flavor of InsertMedia* and EjectMedia methods. Happy to PR either one.

mmlb:
closes #418

### DIFF
--- a/redfish/virtualmedia.go
+++ b/redfish/virtualmedia.go
@@ -270,21 +270,11 @@ func (virtualmedia *VirtualMedia) EjectMedia() error {
 
 // InsertMedia sends a request to insert virtual media.
 func (virtualmedia *VirtualMedia) InsertMedia(image string, inserted, writeProtected bool) error {
-	if !virtualmedia.SupportsMediaInsert {
-		return errors.New("redfish service does not support VirtualMedia.InsertMedia calls")
-	}
-
-	t := struct {
-		Image          string
-		Inserted       bool
-		WriteProtected bool
-	}{
+	return virtualmedia.InsertMediaConfig(VirtualMediaConfig{
 		Image:          image,
 		Inserted:       inserted,
 		WriteProtected: writeProtected,
-	}
-
-	return virtualmedia.Post(virtualmedia.insertMediaTarget, t)
+	})
 }
 
 // VirtualMediaConfig is an struct used to pass config data to build the HTTP body when inserting media
--- a/redfish/virtualmedia.go
+++ b/redfish/virtualmedia.go
@@ -7,6 +7,7 @@ package redfish
 import (
 	"encoding/json"
 	"errors"
+	"net/http"
 	"reflect"
 
 	"github.com/stmcginnis/gofish/common"
@@ -269,7 +270,7 @@ func (virtualmedia *VirtualMedia) EjectMedia() error {
 }
 
 // InsertMedia sends a request to insert virtual media.
-func (virtualmedia *VirtualMedia) InsertMedia(image string, inserted, writeProtected bool) error {
+func (virtualmedia *VirtualMedia) InsertMedia(image string, inserted, writeProtected bool) (*Task, error) {
 	return virtualmedia.InsertMediaConfig(VirtualMediaConfig{
 		Image:          image,
 		Inserted:       inserted,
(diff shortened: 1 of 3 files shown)


mmlb:
@stmcginnis let me know how you feel about this first draft

stmcginnis:
Just a quick note to say this looks good so far. A little busy on other things at the moment, but will get some time to review this soon. Thanks for working on this!

redfish/virtualmedia.go
```diff
@@ -300,12 +291,26 @@ type VirtualMediaConfig struct {
...
+		return nil, errors.New("redfish service does not support VirtualMedia.InsertMedia calls")
+	}
+
+	resp, err := virtualmedia.PostWithResponse(virtualmedia.insertMediaTarget, config)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusAccepted {
```
stmcginnis, line 305: I'm a little worried some vendor implementations might just return a 200 here, just based on some of the oddities I've seen reported with some implementations. But I'm not suggesting this be changed, just noting it so if we run in to a report of an issue there is a breadcrumb here. :D

stmcginnis (commented):
One small tweak if you wouldn't mind. This is great. I think we may want to replicate this pattern elsewhere any time there is an async operation like this that returns a task Location.

I've seen at least one vendor that returns the Location in the body instead of the header. But I can't find an example of that now. But at least with something like this in place it will be easy enough to add a check and grab the task location if it's available anywhere in the response.

redfish/virtualmedia.go
mmlb: Would a TODO or similar comment make more sense than history-in-issues?

mmlb:
> One small tweak if you wouldn't mind. This is great. I think we may want to replicate this pattern elsewhere any time there is an async operation like this that returns a task Location.
> 
> I've seen at least one vendor that returns the Location in the body instead of the header. But I can't find an example of that now. But at least with something like this in place it will be easy enough to add a check and grab the task location if it's available anywhere in the response.

I'm missing what the small tweak you had in mind is. Is it reading from body along with link header? Or also looking at 200 responses?

redfish/virtualmedia.go
stmcginnis: Yeah, not a bad idea!

stmcginnis (approved):
Oops, I think I was going to suggest also looking in the body. But since I can't find an example of what I've seen returned from other vendors, I don't think we need to wait on that. If/when it's encountered again we can easily expand this without changing the way library users interact with the call. So this is good!

I'll wait to merge in case you want to add that TODO, but I don't think that's a blocker and if I don't hear back after awhile I'll just merge it. Thanks!

stmcginnis:
OK, I found some notes at least. Just capturing them here.

I see in some private code I needed to do three things to get it to reliably work with all vendors.

1. `resp.Headers.Get("Location")` would contain the link to the task (main case)
2. `resp.Headers.Get("Location")` would return a link to a `TaskMonitor` instead of a `Task`. Just keying off of `TaskMonitor` being in the string was enough to differentiate them on all platforms I had used
3. If no header, then the `resp.Body` would actually contain the full `TaskMonitor` output that could be directly marshalled into a `TaskMonitor` struct.

In the cases of the `TaskMonitor` results I would grab the `TaskLink` to be able to consistently work with the actual `Task`.

mmlb:
Alright I've added the todos and will keep an eye our for headers vs body. I'll try out the better refresh, thanks!

stmcginnis (approved):
Thanks!

PR #423, opened 2025-05-14
Add support for ActionInfo to common and redfish.VirtualMedia

mmlb:
This PR adds support for fetching and parsing ActionInfo to common and also redfish.VirtualMedia.


mmlb:
@stmcginnis here's an initial attempt. I've only added support to VirtualMedia for now since thats my use case. I can add it to all the other endpoints that would benefit if you want. I'd really like your thoughts on the changes I made to common.ResourceCollection and if common is the appropriate package or if this should live in redfish.

mmlb:
I'll work on the tests, totally misses that! :D

mmlb:
alrighty @stmcginnis I've got some tests in now PTAL

stmcginnis:
Sorry, I went ahead with #434 for the base ActionInfo implementation. Can you rebase and update this so we can pick up the object bits? Let me know if you're busy and I can see if I can help. Thanks!

Issue #425, opened 2025-05-30
Client-side modification of the Service root before creating the Session


dennisklein:
One of my BMCs does not comply with the Redfish standard by not populating the `Links->Sessions->@odata.id` key in the returned Service root (essentially https://github.com/stmcginnis/gofish/issues/162). With this hack I can workaround the issue: https://github.com/dennisklein/gofish/commit/7da1d0fe822aedc9b253b287312415a6e383a30e

I am looking for a better way to solve this *client-side* (The node is actually an older Huawei node on which I cannot update the BMC FW). As far as I understand, I cannot use the approach used in the packages in the `oem` directory in this case because the `gofish.ConnectContext()` API atomically exposes retrieval of the Service root and usage of the SessionService. Would you be open for adding a new API or an API change that would allow me to apply a vendor-specific patch on the Service root object *before* the Session is created?

Idea A - Two-Step API:
```golang
client, err := gofish.SetupClientWithConfig(ctx, config)
if client.Service.Vendor == "Huawei" {
  huaweiService, err := myhuaweipatch.FromService(client.Service)
}
err := client.ConnectContext(ctx)
defer client.Logout()
```

Idea B - Passing a patch lambda:
```golang
client, err := gofish.ConnectContextWithServicePatch(ctx, config, func(*gofish.Service service) {
  if service.Vendor == "Huawei" {
    huaweiService, err := myhuaweipatch.FromService(client.Service)
  }
})
defer client.Logout()
```

What do you think?

PR #426, opened 2025-06-02
Patchable ServiceRoot

dennisklein:
An first attempt at solving #425.

Made `SetupClientWithConfig()` public and added a setter for `APIClient.auth`. You can see an example usage in the added test case.

However, this is not yet finished, see my inline comment.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -8,8 +8,11 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -200,3 +204,175 @@ func TestClientRunRawRequestNoURL(t *testing.T) {
...
+	_, isHuawei = result["Huawei"]
+	if !isHuawei {
+		return nil
+	}
+
+	// This code would be called from an external package
+	// => can only use public API!
+	if s.sessions == "" {
+		s.sessions = "/redfish/v1/SessionService/Sessions"
+	}
```
dennisklein, line 300: Not sure about how to solve this one.

This feels too specific if we added now a Setter for `Service.sessions`.

Another approach would be to have a hook just before the decoding of the JSON response.

What do you think?

dennisklein:
Found a different solution that does not require any change in gofish :) See #425

dennisklein:
> Glad you figured it out! Any details you can share in case anyone else comes along with the same issue?

https://gist.github.com/dennisklein/518dbe9a3c01947a5a6fb562566fe69b

PR #435, opened 2025-06-21
fix: TrustedComponent TPM is a json object

bbrown-cw:
Redfish TrustedComponents can have a `TPM` object that looks like this:

```json
  "TPM": {
    "CapabilitiesVendorID": "IFX",
    "HardwareInterfaceVendorID": "0x15D1"
  },
```

Currently, the gofish TrustedComponent type has a `string` struct field for TPMs, and when using the gofish `ListReferencedTrustedComponents`, the Unmarshal fails.

There is a custom TPM struct that we should use as the type instead

```go
// TPM shall contain TPM-specific information for a trusted component.
type TPM struct {
	// CapabilitiesVendorID shall contain an ASCII string of the 4-byte TCG-defined 'TPM Capabilities Vendor ID' for
	// this trusted component.
	CapabilitiesVendorID string
	// HardwareInterfaceVendorID shall contain the TCG-defined 'TPM Hardware Interface Vendor ID' for this trusted
	// component with the most significant byte shown first.
	HardwareInterfaceVendorID string
}

```

### DIFF
--- a/redfish/trustedcomponent.go
+++ b/redfish/trustedcomponent.go
@@ -106,7 +106,7 @@ type TrustedComponent struct {
 	Status common.Status
 	// TPM shall contain TPM-specific information for this trusted component. This property shall only be present for
 	// TCG-defined TPM trusted components.
-	TPM json.RawMessage `json:"TPM,omitempty"`
+	TPM TPM
 	// TrustedComponentType shall contain the type of trusted component.
 	TrustedComponentType TrustedComponentType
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Thanks - great catch!

Not going to hold on this, but with the `This property shall only be present` comment in the spec, I wonder if it might be better to do:

```go
TPM *TPM `json:",omitempty"`
```

Just noting that here if someone wants to do a follow up.

PR #439, opened 2025-07-11
Handle etag properly on CreateAccount call

joeyberkovitz:
I discovered that BMCs send a different ETag on the `/AccountService/Accounts` endpoint, which is used to both list accounts and create a new account. On some BMCs, the ETag is the same for both `/AccountService` and `/AccountService/Accounts`

As a result, `CreateAccount` fails if ETag is properly validated on a BMC. This change adjusts the request to retrieve the proper ETag before creating an account if ETag is enabled

### DIFF
--- a/common/entity.go
+++ b/common/entity.go
@@ -63,6 +63,11 @@ func (e *Entity) DisableEtagMatch(b bool) {
 	e.disableEtagMatch = b
 }
 
+// IsEtagMatchDisabled indicates if etag matching is enabled on this entity
+func (e *Entity) IsEtagMatchDisabled() bool {
+	return e.disableEtagMatch
+}
+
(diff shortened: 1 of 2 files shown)


stmcginnis:
Thanks @joeyberkovitz! Can you share which vendors you have validated this against?

joeyberkovitz:
let me contact you privately, I want to be sensitive about sharing our vendor information publicly.

stmcginnis:
Oh, sorry! I don't want you to have to share any sensitive information. Let me rephrase the question then.

My main concern is whether this has been tested against more than one vendor's implementation. I want to make sure this isn't just a quirk with a specific vendor that is going to cause problems when running against a different one.

So I guess I just need to know - has this been tested against more than a single vendor?

PR #440, opened 2025-07-18
Complete support for Thermal.v1_7_1 schema

Muyk33rus:
Improved handling of fan and temperature sensors data OEM extensions support for vendor-specific implementations

### DIFF
--- a/redfish/thermal.go
+++ b/redfish/thermal.go
@@ -15,91 +15,75 @@ import (
 type ReadingUnits string
 
 const (
-
-	// RPMReadingUnits Indicates that the fan reading and thresholds are
-	// measured in rotations per minute.
+	// RPMReadingUnits indicates that the fan reading and thresholds are measured in rotations per minute.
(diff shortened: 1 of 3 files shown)


redfish/thermal.go
```diff
@@ -24,82 +24,95 @@ const (
...
-	// ReadingUnits shall contain the units in which the fan reading and thresholds are measured.
-	ReadingUnits ReadingUnits
-	// Redundancy shall contain an array of links to the redundancy groups to which this fan belongs.
-	redundancy []string
-	// RedundancyCount is the number of Redundancy items.
-	RedundancyCount int `json:"Redundancy@odata.count"`
-	// RelatedItem shall contain an array of links to resources or objects that this fan services.
+
+	// Assembly contains a link to a resource of type Assembly
+	Assembly string `json:"Assembly"`
```
stmcginnis, line 32: This shouldn't be made public. Links like this should stay private, with `Assembly()` used to access the Assembly object.

PR #441, opened 2025-07-23
Allow non standard transports to reuse connections.

thespags:
If we want to create a custom transport layer, for instance, one that may cache request responses, then we can't get the benefit of reusing existing connections.

Is there concern this is misleading in case you create a custom transport layer that doesn't DisableKeepAlive?


stmcginnis (approved):
Thanks, this seems reasonable.

PR #452, opened 2025-08-07
feat: add WithContext function to client

joeyberkovitz:
right now, the context setup is a bit clunky. I think the ideal situation would be providing a context argument on every function call to allow for different context for different requests.

I understand that it's not really feasible given that it would be essentially be a full rewrite. Instead I propose adding `WithContext` to the `APIClient` to allow for creating a copied client, using the same auth and everything, just with a different context.

### DIFF
--- a/client.go
+++ b/client.go
@@ -268,6 +268,13 @@ func (c *APIClient) GetService() *Service {
 	return c.Service
 }
 
+// WithContext returns a copy of the client using the provided context
+func (c *APIClient) WithContext(ctx context.Context) *APIClient {
+	newClient := *c
+	newClient.ctx = ctx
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Seems reasonable. I could see this being useful. Thanks!

joeyberkovitz:
I can write a unit test later to validate this, but I have a feeling it's not an issue because WithContext copies the entire struct and just changes the context in the copied struct.

stmcginnis:
Nevermind, you are correct. I was thinking we were just grabbing an updated reference, but now I see you're creating a copy. So we should be all good. Sorry for the confusion.

PR #458, opened 2025-09-10
feat(redfish): Add function to handle async SPDMGetSignedMeasurements

dukovac:
Many implementations of the `SPDMGetSignedMeasurements` action returns a Task, the requester then polls the Task URI until the Task is complete. Although this response is not explicitly listed in the spec https://www.dmtf.org/sites/default/files/standards/documents/DSP2046_2025.2.html#componentintegrity-131 in theory, every non-GET operation can result in a task (more discussion https://redfishforum.com/thread/735/component-integrity-action-synchronous-asynchronous).

### DIFF
--- a/redfish/componentintegrity.go
+++ b/redfish/componentintegrity.go
@@ -6,6 +6,7 @@ package redfish
 
 import (
 	"encoding/json"
+	"net/http"
 
 	"github.com/stmcginnis/gofish/common"
 )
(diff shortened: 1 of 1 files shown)


stmcginnis:
Thanks @dukovac. I do want to get this pattern in across all operations like this.

We're still a v0.x release, so it's possible to make breaking changes. I try to avoid it when possible, but sometimes it makes more sense. Rather than adding a second method for the same thing, we've already [set the precendent](https://github.com/stmcginnis/gofish/pull/419/files#diff-a261f5d9e8da09919a58e2e8af249b29b8650a232cfb1f349f655c534cc79cae) of just updating the existing call's signature to return the task if available. Could you update this to just change the existing `SPDMGetSignedMeasurements`?

stmcginnis (approved):
Thanks, this looks good to me.

PR #460, opened 2025-09-17
Add Reset action to the PowerSupply type

smiller248:
`PowerSupplyUnit` has a reset action but `PowerSupply` (whose name matches the actual underlying redfish type) doesn't. This rectifies that gap. Thanks!


stmcginnis (approved):
Thanks!

PR #470, opened 2025-11-04
common/types: handle fallback case for malformed operations entries

### LINKED ISSUE #469: drive: operations/associated task
:wave: We've got some new SMC BMC upgrades that are causing the JSON unmarshal for drives to fail due to the `Operations/AssociatedTask` now appears to be an empty array.

A snippet from the redfish call to `/redfish/v1/Chassis/NVMeSSD.0.Group.0.StorageBackplane/Drives/Disk.Bay.0`:

```
  "@odata.type": "#Drive.v1_20_1.Drive",
  "@odata.id": "/redfish/v1/Chassis/NVMeSSD.0.Group.0.StorageBackplane/Drives/Disk.Bay.0",
  "Operations": [
    {
      "AssociatedTask": [],
      "OperationName": null,
      "PercentageComplete": 0
    }
  ],
```

The _easiest_ thing I can think of is to `s/string/any` [here](https://github.com/stmcginnis/gofish/blob/52d3b81c7ff5877c5bd908e551e43053a4f116c2/common/types.go#L556), but I'm not sure if that's acceptable. In the DMTF/Redfish repo, the openapi spec does not describe a type for the `AssociatedTask` field, though it does for `OperationName` and `PercentageComplete`.

The unmarshal error:

```
failed to retrieve some items: [{"link":"/redfish/v1/Chassis/NVMeSSD.0.Group.0.StorageBackplane/Drives/Disk.Bay.0","error":"json: cannot unmarshalarray into Go struct field Operations.temp.Operations.AssociatedTask of type string"},...]
```

For reference, from an older BMC version on the same hardware:

```
  "@odata.type": "#Drive.v1_9_0.Drive",
  "@odata.id": "/redfish/v1/Chassis/NVMeSSD.0.Group.0.StorageBackplane/Drives/Disk.Bay.0",
```

Simply omits the `Operations[]`.

Is there an appetite to consider changing this field type in gofish, or is there perhaps another suggestion?

Matt1360:
Unfortunately, with some eject/insert/blink/unblink, I'm unable to get a pulse on what the BMC will actually present in the `AssociatedTask[]`. I've taken a _guess_ that it might be a `[]Link`.

I've tested this with our nodes, on both the seemingly compliant (missing `Operations`) older version, and newer version of the BMC.

Fixes #469

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -551,14 +551,63 @@ type ResourceCollection struct {
 
 // Operations shall describe a currently running operation on the resource.
 type Operations struct {
-	// AssociatedTask shall be a reference to a resource of type Task that
+	// associatedTask shall be a reference to a resource of type Task that
 	// represents the task associated with the operation.
-	AssociatedTask string
+	associatedTask Link
 	// OperationName shall be a string of the name of the operation.
 	OperationName string
 	// PercentageComplete shall be an integer of the percentage of the
 	// operation that has been completed.
 	PercentageComplete int
+}
+
+// UnmarshalJSON unmarshals an Operations object from the raw JSON.
+func (ops *Operations) UnmarshalJSON(b []byte) error {
+	type temp Operations
+	type t1 struct {
+		temp
+		AssociatedTask Link
+	}
+	var t t1
+
+	err := json.Unmarshal(b, &t)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


stmcginnis:
@joeyberkovitz would you have time to take a look at this and let me know what you think, related to your TaskManager changes?

PR #482, opened 2025-11-28
expand support for entities and collections

joeyberkovitz:
Adds expand support for both entities and collections

Provides an option via the client to auto-enable `$expand` if supported on collections

### DIFF
--- a/common/entity.go
+++ b/common/entity.go
@@ -27,9 +27,9 @@ type Entity struct {
 	// ODataEtag contains either the entity's declared ETag or the ETag header
 	// when fetching the object to control updates and prevent conflicts between
 	// concurrent modifications
-	ODataEtag string `json:"@odata.etag"`
+	ODataEtag string `json:"@odata.etag,omitempty"`
 
-	ExtendedInfo []MessageExtendedInfo `json:"@Message.ExtendedInfo"`
+	ExtendedInfo []MessageExtendedInfo `json:"@Message.ExtendedInfo,omitempty"`
 
 	// stripEtagQuotes removes surrounding quotes of etag used in If-Match header
 	// Only use for vendor implementations where If-Match only matches unquoted etag


common/collection.go
```diff
@@ -156,13 +209,31 @@ func GetCollectionObjects[T any, PT interface {
...
+			// if the entity has any ExtendedInfo, we assume it's an error
+			var err error
+			extendedInfo := entity.GetExtendedInfo()
+			if len(extendedInfo) > 0 {
+				errE := &Error{}
+				for _, info := range extendedInfo {
+					errE.ExtendedInfos = append(errE.ExtendedInfos, ErrExtendedInfo(info))
+				}
+				err = errE
+			}
```
joeyberkovitz: spec wasn't entirely clear to me here - does each entity get an extended info or the collection?

joeyberkovitz: I think what I have is correct https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.html#the-expand-query-parameter:~:text=The%20following%20example%20response%20shows%20a%20partial%20failure%20of%20a%20deep%20PATCH%20where%20the%20ETag%20provided%20in%20the%20request%20for%20the%20Role%20resource%20named%20ReadOnlyRestricted%20was%20incorrect%3A

spec also says:

> If a service cannot return the payload corresponding to an individual [member](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.html#member) of a resource collection, but can return other members, it should:
> 
> Return the HTTP [200 OK](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.html#status-200) status code.
> Return the [@odata.id](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.html#resource-identifier-odataid-property) property for the member that it cannot expand in the Members array.
> Return [extended information](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.22.1.html#extended-information) with the member indicating the reason that member was not returned, such as when a provider internal to the service returns an error or times out.

PR #484, opened 2025-12-09
fix: remove shadowed embedded struct fields

joeyberkovitz:


joeyberkovitz:
@stmcginnis - draft PR but there are two issues here:

- I added `@odata.etag` to `common.Entity` because it's used on everything and needed for expand. I failed to realize that it's present on some structs as well
- part 2: the `Update` func detects that a read-only field like `@odata.etag` is being modified and returns an error on updates - can we just make this silently skip any fields that were modified?

PR #492, opened 2026-01-14
add OEM fields for BootProgress struct

glee-cw:
@stmcginnis - added the `Oem` and `OemLastState` field to to the `BootProgress` struct to be in line with the redfish spec https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2025.3.pdf

### DIFF
--- a/redfish/computersystem.go
+++ b/redfish/computersystem.go
@@ -571,6 +571,12 @@ type BootProgress struct {
 	LastState BootProgressTypes
 	// LastStateTime shall contain the date and time when the last boot state was updated.
 	LastStateTime string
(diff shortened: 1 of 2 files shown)


glee-cw:
> This seems like something better to surface in one of the OEM object implementations?

@stmcginnis Would you be able to elaborate as to what that would look like?

joeyberkovitz:
to be clear, these two fields are explicitly defined in the spec https://www.dmtf.org/sites/default/files/standards/documents/DSP0268_2025.2.html#coolingloop-103:~:text=BootProgress%20(v1.13%2B)%20%7B

any reason to not support them?

stmcginnis (approved):

PR #509, opened 2026-02-27
Update pciedevice.go

thespags:
My previous fix https://github.com/stmcginnis/gofish/pull/443 got removed. Adding back. Nvidia HGX for whatever reason is incorrectly using [] instead of null. This handles the vendor one-off.


stmcginnis (approved):
Oops, sorry about that. Thanks for redoing the fix.

PR #513, opened 2026-03-07
schemas: preserve collection member order

jmejco:
`GetCollectionObjects` and `GetObjects` fetch collection members concurrently but return them in HTTP response completion order, not the order listed in the collection’s `Members` array. This makes slice ordering dependent on hardware and timing.

### Background

On Dell iDRAC10, the VirtualMedia collection exposes two slots. Only `VirtualMedia/1` is bootable via the standard `BootSourceOverrideTarget=Cd`, and consumers like `bmclib` typically select the first returned slot. Because gofish returned collection members in completion order, `ComputerSystem.VirtualMedia()` was non-deterministic and frequently returned `VirtualMedia/2` first, causing ISO mounts to land on the non-bootable slot and the host to fail to boot.

### Change

`GetCollectionObjects` and `GetObjects` now preserve member order by writing results into a preallocated slice at the original member index (link → index map). Results are compacted to a dense slice before returning. Concurrency is unchanged.

A helper (`collectMemberLinks`) walks collection pages to collect member links in server order before dispatching concurrent fetches.

### Testing

* Hardware repro against Dell R470 iDRAC10 calling `systems[0].VirtualMedia()` 20 times:

  * main: `VirtualMedia/1` first 9/20 (wrong slot ~55%)
  * this branch: `VirtualMedia/1` first 20/20
* Added unit test `TestGetCollectionObjectsOrder` (members `/3, /1, /2` → results match that order).

### DIFF
--- a/schemas/collection.go
+++ b/schemas/collection.go
@@ -193,65 +193,129 @@ func CollectResourceCollection[T any, PT interface {
 	wg.Wait()
 }
 
+// extendedInfoToError converts a slice of Message into a *Error.
+func extendedInfoToError(ext []Message) *Error {
+	errE := &Error{}
+	for i := range ext {
+		errE.ExtendedInfos = append(errE.ExtendedInfos, ErrExtendedInfo(ext[i]))
+	}
+	return errE
+}
+
+// collectMemberLinks walks all pages of a typed collection starting at uri,
+// returning the ordered list of member @odata.id links. queryOpts are applied
+// only to the first page; subsequent pages follow MembersNextLink verbatim.
+// Inline-expanded members that carry ExtendedInfo are recorded as errors and
+// excluded from the returned links. Inline-expanded members with a valid Id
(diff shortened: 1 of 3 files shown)


jmejco:
> This seems more like a usage error? Nothing should be assuming conditions based on slice ordering. If you need a certain instance you should check which one you are using, right?

Thank you for your response. I agree that consumers should not assume conditions based on ordering.

For some context, this came up while we were using EKS Anywhere, which uses bmclib under the hood to ISO boot Dell BMC nodes. During testing, we found that about 60% of the time the nodes would fail to one-time boot. Tracing it back, the issue was that gofish was returning the virtual media list in a different order from one call to the next, even though raw Redfish was consistently returning the same order. I did look at fixes at other layers as well, like adding selection logic in bmclib or allowing the VM slot to be passed in from EKS Anywhere.

I opened the PR here because preserving the collection order returned by the service seems like the right client-layer behavior, while any selection logic on top of that can still remain with consumers like bmclib.

PR #530, opened 2026-05-26
Workaround for Lenovo VirtualMedia handling

### LINKED ISSUE #529: VirtualMedia mounting does not work on Lenovo ThinkSystem SR630 v2 BMC.
## Issue

When attempting to mount an ISO image as VirtualMedia on this server, the following error occurs:
```text
2026/05/26 11:23:28 error inserting media: unable to execute request, no target provided
```

## Root Cause

Further investigation revealed that the endpoint, parameters, and request type for VirtualMedia mounting on ThinkSystem SR630 v2 differ from the standard Redfish [VirtualMedia.v1_6_5.json](https://redfish.dmtf.org/schemas/v1/VirtualMedia.v1_6_5.json) schema.

The endpoint does not have Actions, and a PATCH request must be used instead of POST.

## Expected Behavior

The request body should be formatted as follows:
```json
{
    "Image": "http://192.168.1.2/Core-current.iso",
    "Inserted": true,
    "WriteProtected": true,
    "UserName": "test",
    "Password": "PASSW0RD"
}
```

## Additional Information

Detailed description is available in the Lenovo XCC REST API documentation:
https://pubs.lenovo.com/xcc-restapi/resource_managers_one_virtualmedia

## Environment

- **Server**: Lenovo ThinkSystem SR630 v2
- **BMC/XCC Version**: 1.89 (Build ID: AFBT20X)

stmcginnis:
Some systems, like some Lenovo ones, do not support the Insert and Eject VirtualMedia actions and instead rely on PATCHing to perform these actions. This adds a workaround to the calls to attempt this approach rather than failing right away if if the actions are not supported.

Closes: #529

### DIFF
--- a/schemas/virtualmedia.go
+++ b/schemas/virtualmedia.go
@@ -290,7 +290,13 @@ func (v *VirtualMedia) EjectMediaActionInfo() (*ActionInfo, error) {
 // If TaskMonitorInfo is not nil it can be used to monitor async tasks.
 func (v *VirtualMedia) EjectMedia() (*TaskMonitorInfo, error) {
 	if !v.SupportsMediaEject {
-		return nil, errors.New("redfish service does not support VirtualMedia.EjectMedia calls")
+		// Some implementations (e.g. Lenovo XCC) do not expose the standard
+		// VirtualMedia.EjectMedia action; fall back to PATCHing the resource.
+		payload := struct {
+			Inserted       bool
+			WriteProtected bool
+		}{}
+		return nil, v.Patch(v.ODataID, payload)
 	}
 
 	payload := make(map[string]any)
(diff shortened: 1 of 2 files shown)


stmcginnis:
@nuxster - would you be able to try this change against your system to see if it addressing your issue?

nuxster:
I will test it soon! Thanks!

nuxster:
@stmcginnis It works great, I was able to mount several images at once! Thank you so much for your work!

stmcginnis:
Awesome! To be clear - I do still think this is something Lenovo should address on their end. But pretty trivial workaround we can do to make lives easier, so I'll merge this in. Thanks for verifying it!

PR #536, opened 2026-06-26
Configurable request concurrency

### LINKED ISSUE #535: Allow concurrency to be configurable for fetching collections
Hi,

Would you be open to accept a change to allow the concurrency of requests to be configurable?
https://github.com/stmcginnis/gofish/blob/01dc1f8aaa368970f684ec95dac1acef50e9a1e3/schemas/collection.go#L178-L179

We've tried this for our system and 3 is just way to low.
When retrieving information for the whole system, it takes a solid 1m to get everything.

And if I modify this to 16, the total wall time becomes ~10s.

While I understand the rational behind this section, I think it's best to allow the user to decide the level of concurrency they want and if it'll saturate the system or not.

darwin67:
resolves #535 

This allows request concurrency to be configurable from the caller, but still keep the default to 3, like how it is today.

### DIFF
--- a/schemas/collection.go
+++ b/schemas/collection.go
@@ -151,9 +151,10 @@ func CollectListGeneric[T any, PT interface {
 		}
 	}
 
-	CollectResourceCollection(get, collection.Members, queryOpts...)
+	collectionQuery := BuildQueryGroup(c, queryOpts...).QueryCollection
+	collectResourceCollection(get, collection.Members, collectionQuery.collectionRequestConcurrency, queryOpts...)
 	if collection.MembersNextLink != "" {
-		err := CollectListGeneric(get, c, collection.MembersNextLink)
+		err := CollectListGeneric(get, c, collection.MembersNextLink, queryOpts...)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


stmcginnis:
Thanks for working on this, but is there a reason not to use [MaxConcurrentRequests](https://github.com/stmcginnis/gofish/blob/main/client.go#L100) for this? If there is a reason to have a separate knob, then there needs to be a better or more obvious way for the two values to work together so [it still doesn't get throttled](https://github.com/stmcginnis/gofish/blob/main/client.go#L128).

There also isn't an easy public API for configuring a client's collection concurrency. Can you provide an example to show how you see this working?

darwin67:
oh, I think I simply missed `MaxConcurrentRequest`.

Let me give that a try and report back if it solves what I'm aiming for.

darwin67:
Just tried with `MaxConcurrentRequest` and this works.
Retrieving the full device info takes like ~10s which seems mostly similar

stmcginnis:
> Just tried with `MaxConcurrentRequest` and this works. Retrieving the full device info takes like ~10s which seems mostly similar

Awesome, thanks @darwin67!

PR #542, opened 2026-07-08
Restore manager OemLinks

Keruspe:
Since the switch to generated manager.go, the OemLinks field got lost.
Restore it, and while at it, also parse some of it for Dell.
I haven't hooked it up in the generator as I'm not sure exactly in which shape it is as it does not seem to be generating the current state of the files


stmcginnis (approved):
This is kind of an odd one since it's nested under the links. So not something the generator would handle for us. Thanks!

PR #546, opened 2026-07-21
Handle numeric AccountService ETags

### LINKED ISSUE #545: AccountService parsing fails when iDRAC returns numeric @odata.etag
## Problem

`AccountService()` fails when Dell iDRAC9 returns `@odata.etag` as a JSON number instead of a string.

## Environment

- GoFish: `v0.23.0`
- iDRAC9 firmware: `4.10.10.10`
- Redfish: `1.6.0`
- AccountService: `v1_3_1`

## Response

The request returns HTTP 200, but the response body contains a numeric ETag:

```json
{
  "@odata.id": "/redfish/v1/AccountService",
  "@odata.etag": 87422082150,
  "Id": "RemoteAccountService",
  "Name": "Account Service"
}
```

GoFish returns:

```text
json: cannot unmarshal number into Go struct field .temp.Entity.@odata.etag of type string
```

## Expected behavior

Please allow `@odata.etag` to be decoded from either a JSON string or number, while keeping its Go representation as a string. Missing or null values should continue to be accepted.

Gintoms:
Some Redfish implementations return `AccountService.@odata.etag` as a JSON number. This currently fails because `Entity.ODataEtag` is a string.

Accept string, number, null, or a missing value for AccountService ETags and normalize numbers to their string representation. Other JSON types remain invalid, and the public API is unchanged.

Tests:
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go vet ./...`

Fixes #545

### DIFF
--- a/schemas/accountservice.go
+++ b/schemas/accountservice.go
@@ -8,6 +8,7 @@
 package schemas
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 )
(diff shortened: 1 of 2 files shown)


Gintoms:
Closing based on the maintainer feedback in #545. The numeric ETag is not valid according to the Redfish schema, so we will address this by upgrading the iDRAC firmware and reporting it to Dell if the behavior persists, rather than adding a client-side workaround.

PR #547, opened 2026-07-29
Add PatchWithTask and UpdateBiosAttributesApplyAtWithTask

### LINKED ISSUE #472: Task Support
See #469 and #246

I wanted to start a discussion and come to a conclusion before PRing anything

Many endpoints can return a Task Monitor URL with an optional Task body. The Task Monitor is an entirely opaque URL that's just polled until the status code changes from 202 to something else, at which point it returns the same response as the request would've returned had it been synchronous (a Location header and/or an HTTP body).

The spec allows for the task monitor body to be a Task, although the body may also be entirely empty, so we shouldn't really rely on it existing. 

It would be nice if gofish had more generic support for task monitors:
- check if the body of a response is actually a task (I believe it's usually minimal with just the task ID and a submitted message)
- regardless, if you get a 202 and a location, that location is a task monitor URL

Then TBD how gofish handles this. Some thoughts: 
- gofish can handle all of the logic for polling the monitor and then when done returning the contents of the location. 
  - I think this is suboptimal mainly because the client might want to be responsible for the waiting
  - probably still makes sense to provide a function to handle the polling
- gofish simply returns the task monitor URL and optional task with a nil result upon 202. Task comes from the action body and monitor from the response header
  - SPDM today kind of does this, but not in a fully spec compliant way. It queries the task monitor URL to try to get the task out of it, but the response doesn't always have a task and is still spec compliant

breml:
Based on the discussion in #472, this PR adds `PatchWithTask` and `UpdateBiosAttributesApplyAtWithTask`.

Since `UpdateBiosAttributesApplyAt` is the only place I observed a BMC to return a task monitor on a PATCH method, this PR only adds `...WithTask` for this function. That being said, other cases can be added later on pretty easily.

Closes: #472

### DIFF
--- a/schemas/bios.go
+++ b/schemas/bios.go
@@ -176,6 +176,13 @@ func (bi *Bios) AllowedAttributeUpdateApplyTimes() []SettingsApplyTime {
 
 // UpdateBiosAttributesApplyAt is used to update attribute values and set apply time together
 func (bi *Bios) UpdateBiosAttributesApplyAt(attrs SettingsAttributes, applyTime SettingsApplyTime) error {
+	_, err := bi.UpdateBiosAttributesApplyAtWithTask(attrs, applyTime)
+	return err
+}
+
+// UpdateBiosAttributesApplyAtWithTask is used to update attribute values and set apply time
+// together. It returns a TaskMonitorInfo when the service processes the update asynchronously.
+func (bi *Bios) UpdateBiosAttributesApplyAtWithTask(attrs SettingsAttributes, applyTime SettingsApplyTime) (*TaskMonitorInfo, error) {
 	payload := make(map[string]any)
 
 	// Get a representation of the object's original state so we can find what
@@ -183,7 +190,7 @@ func (bi *Bios) UpdateBiosAttributesApplyAt(attrs SettingsAttributes, applyTime 
 	original := new(Bios)
 	err := original.UnmarshalJSON(bi.rawData)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	for key := range attrs {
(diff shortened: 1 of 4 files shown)


schemas/task_monitor.go
```diff
@@ -79,6 +79,50 @@ func postObjectWithTask[T any,
...
+	resp, err := c.PatchWithHeaders(uri, payload, headers)
+
+	// If 412 Precondition Failed, retry without If-Match header.
+	// Some BMC implementations (e.g., Dell iDRAC10) reject If-Match
+	// on PATCH endpoints.
+	if err != nil && Is412PreconditionFailed(err) {
+		DeferredCleanupHTTPResponse(resp)
+		delete(headers, "If-Match")
+		resp, err = c.PatchWithHeaders(uri, payload, headers)
+	}
```
breml, line 98: The implementation of `PatchWithTask` aligns it self to `PostWithTask`. This is why I also preserved the handling of the 412 status code even though, I am not entirely sure, this applies here as well.

schemas/bios.go
```diff
@@ -176,14 +176,21 @@ func (bi *Bios) AllowedAttributeUpdateApplyTimes() []SettingsApplyTime {
 
 // UpdateBiosAttributesApplyAt is used to update attribute values and set apply time together
 func (bi *Bios) UpdateBiosAttributesApplyAt(attrs SettingsAttributes, applyTime SettingsApplyTime) error {
+	_, err := bi.UpdateBiosAttributesApplyAtWithTask(attrs, applyTime)
+	return err
+}
+
+// UpdateBiosAttributesApplyAtWithTask is used to update attribute values and set apply time
+// together. It returns a TaskMonitorInfo when the service processes the update asynchronously.
+func (bi *Bios) UpdateBiosAttributesApplyAtWithTask(attrs SettingsAttributes, applyTime SettingsApplyTime) (*TaskMonitorInfo, error) {
```
breml, line 185: I decided to only add the `UpdateBiosAttributesApplyAtWithTask` and not `UpdateBiosAttributesWithTask`, since I think, the later will by synchronous and therefore will not return a task monitor.

stmcginnis (approved):
Looks good, thanks!

PR #554, opened 2026-08-20
Add Dell iDRAC reset action discovery

tobiasb-dell:
## Summary

- Discover the Dell `DelliDRACCardService` through the manager OEM links.
- Read the `#DelliDRACCardService.iDRACReset` target advertised by the service instead of posting to a fixed URI.
- Add the task-aware `ResetIDRAC` API while retaining `ResetiDRAC` as a compatibility wrapper.
- Validate reset types and add coverage for resource discovery, synchronous responses, asynchronous task monitors, and error cases.

## Motivation

The existing implementation constructs a fixed Dell OEM action URI containing a specific manager identifier. Redfish resources advertise both the related OEM service and its action target, so following those links avoids assumptions about the service path and manager name. It also allows callers to receive the task monitor returned when iDRAC accepts a reset asynchronously.

Dell documents the action on the [`DelliDRACCardService`](https://developer.dell.com/apis/1796449a-cc87-4882-925b-6241fbf1bfea/versions/1.30.xx/whats-new-9350m0) API.

## Compatibility

The existing `ResetiDRAC` method remains available and delegates to the new implementation. Existing reset constants are retained, with their type exported as `IDRACResetType`. Callers that need asynchronous task information can use `ResetIDRAC`.

## Testing

- `go test ./oem/dell`
- `make test`
- `make lint`
- `git diff --check`

### DIFF
--- a/oem/dell/idrac_monitor_test.go
+++ b/oem/dell/idrac_monitor_test.go
@@ -13,12 +13,6 @@ import (
 	"github.com/stmcginnis/gofish/schemas"
 )
 
-func TestIDRACReset(t *testing.T) {
-	// This would require a mock server, for now just test the basic functionality exists
-	// The actual implementation would need integration testing with a real or mocked iDRAC
-	t.Skip("Skipping iDRAC reset test - requires integration testing with real iDRAC or comprehensive mocking")
-}
-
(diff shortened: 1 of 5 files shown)


stmcginnis (approved):
Looks good, thanks!

PR #555, opened 2026-08-20
Add live Dell iDRAC health checks

tobiasb-dell:
## Summary

- Export `IDRACMonitor` and `IDRACMonitorConfig`, and add the conventionally named `NewIDRACMonitor` constructor.
- Replace the cached manager ID check with a live GET against the configured service or manager resource.
- Enforce the configured health-check timeout with the existing `APIClient.WithContext` mechanism.
- Retry only transient network errors, HTTP 408/429/5xx responses, and deadline expirations.
- Add configurable retry intervals and comprehensive tests for health checks, retries, cancellation, and permanent failures.

## Problem

The monitor introduced in #488 creates a timeout context but never uses it. `CheckHealth` currently reports success whenever a previously loaded manager has a non-empty ID, even if iDRAC is no longer reachable. The configuration type is also unexported, so external callers cannot supply non-default settings.

Operation retry detection additionally relies on recursive, case-insensitive matching of error strings. That can misclassify errors and does not distinguish transient Redfish responses from permanent client errors.

## Implementation

The monitor obtains the `APIClient` attached to the selected Redfish resource and performs the health request with:

```go
client.WithContext(requestContext).Get(resource.ODataID)
```

This reuses the existing client request path, including authentication, transport configuration, concurrency limiting, and context cancellation. No generic client methods or interfaces are added.

Health and operation retries now:

- stop immediately for cancellation and permanent errors;
- retry timeouts, common connection failures, HTTP 408/429, and HTTP 5xx;
- apply a context-aware linear retry delay; and
- only consider an automatic reset after transient health failures exhaust all attempts.

## Compatibility and scope

The existing `NewiDRACMonitor` constructor remains as a deprecated compatibility wrapper. The existing reset API is unchanged.

The branch is rebased onto current `main`, including #554. The PR changes only the Dell iDRAC monitor and its tests.

Live health checks require the resource to use the standard `gofish.APIClient`; a clear error is returned for a missing or different client implementation.

## Testing

- `make test`
- `make lint`
- `go test -race ./oem/dell`
- `git diff --check`

### DIFF
--- a/oem/dell/idrac_monitor.go
+++ b/oem/dell/idrac_monitor.go
@@ -8,149 +8,220 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"net/http"
+	"syscall"
 	"time"
 
+	"github.com/stmcginnis/gofish"
 	"github.com/stmcginnis/gofish/schemas"
 )
 
-// iDRACMonitor monitors iDRAC responsiveness and can trigger resets
-type iDRACMonitor struct {
-	service        *schemas.Entity // Using Entity as base for service access
+const (
+	defaultIDRACMonitorTimeout       = 30 * time.Second
+	defaultIDRACMonitorMaxRetries    = 3
+	defaultIDRACMonitorRetryInterval = time.Second
+)
+
+// IDRACMonitor monitors iDRAC responsiveness and can trigger resets.
+type IDRACMonitor struct {
+	service        *schemas.Entity
 	manager        *Manager
 	timeout        time.Duration
 	maxRetries     int
+	retryInterval  time.Duration
 	resetOnTimeout bool
 }
 
-// iDRACMonitorConfig configures the iDRAC monitor
-type iDRACMonitorConfig struct {
-	// Timeout for individual requests
+// IDRACMonitorConfig configures the iDRAC monitor.
+type IDRACMonitorConfig struct {
+	// Timeout is the maximum duration of an individual health check. A value of
+	// zero disables the per-request timeout.
 	Timeout time.Duration
-	// Maximum number of retries before considering iDRAC unresponsive
+	// MaxRetries is the number of retries after the initial attempt.
 	MaxRetries int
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -378,6 +378,16 @@ func (c *APIClient) Get(url string) (*http.Response, error) {
 	return c.GetWithHeaders(url, nil)
 }
 
+// GetWithContext performs a GET request using the provided context.
+func (c *APIClient) GetWithContext(ctx context.Context, url string) (*http.Response, error) {
```
stmcginnis: Is there a reason `APIClient.WithContext()` doesn't work for this use case? (e.g. instead of 

```go
resp, err := getter.GetWithContext(requestContext, resource.ODataID)
```

you would just do

```
resp, err := client.WithContext(requestContext).Get(resource.ODataID)
```
)

A lot of this was defined before passing in a `context` was really well established as a common pattern, so it's been difficult to retrofit something like that. We already have a lot of layers in the API client calls to add headers and other things. The `WithContext` method was added to try to give the ability to have cancellable context calls without changing signatures or adding yet more indirection.

tobiasb-dell: You're right: `APIClient.WithContext()` works for this use case, and I missed the existing helper.

I rebased the branch onto current `main` and changed the health check to use:

```go
client.WithContext(requestContext).Get(resource.ODataID)
```

Because `Entity.GetClient()` returns `schemas.Client`, the monitor first type-asserts it to `*gofish.APIClient`. I removed `GetWithContext` and all generic client-layer changes, so the PR now only touches the Dell monitor and its tests. The tests use a real `APIClient` against an `httptest` server to verify timeout cancellation through the existing path.

Thanks for pointing this out.

stmcginnis: Thanks! I wanted to make sure there wasn't a major shortcoming there that I was missing. It's not the best pattern, but it's worked so far to avoid changing the signature of everything. ;)

stmcginnis (approved):

PR #569, opened 2026-09-09
schemas: make Boot.HttpBootURI a pointer so it can be cleared

mcanevet:
HttpBootUri is nullable in the schema, but MapType's string case always returns isPointer=false, so it never got the same nullable->pointer treatment as the int/float cases. Result: omitempty drops an explicit empty value the same as "not set", so there's no way to PATCH the URI back to empty once it's been set.

Special-cased HttpBootUri by name (same pattern as UUID/ID/Description/Name) instead of making all nullable strings pointers - Boot has ~15 other nullable string fields, and a blanket fix would flip things like ComputerSystem.SerialNumber/Manufacturer/Model to *string too (read as plain strings elsewhere, would break callers), plus incorrectly turn BootOrder []string into []*string.

Regenerated against DMTF 2025.4, the version already reflected in the file, to avoid pulling in the unrelated 2026.1 schema bump that's landed since.


stmcginnis (approved):
Thanks for the thorough update!

PR #570, opened 2026-09-10
schemas: add exact-attribute Bios update variants

mcanevet:
## Summary

`Bios.UpdateBiosAttributesApplyAt`/`UpdateBiosAttributesApplyAtWithTask` diff the requested attributes against the resource's *currently reported* state and silently omit anything that already matches - a reasonable default for writing back a subset of a freshly-read attribute map, but it leaves no way for a caller to force an attribute into the PATCH when it's needed there regardless of the current value.

That matters for a caller resubmitting a previously-read pending/staged attribute set after recovering from a conflict (e.g. a vendor that only allows one pending settings job at a time, and rejects a second PATCH until the first is applied or cancelled): an attribute that happens to already match the currently-applied value - because the caller is reverting it back to what it was - gets silently dropped from the resubmission instead of being explicitly (re)included, corrupting what actually ends up staged.

This PR adds `UpdateBiosAttributesExactApplyAt`/`UpdateBiosAttributesExactApplyAtWithTask`/`UpdateBiosAttributesExact`, which behave exactly like their non-`Exact` counterparts except every key in the requested attributes is always included in the PATCH. The shared PATCH-sending logic (ETag fetch, apply-time header, `PatchWithTask`) is factored out into a new `patchAttributesApplyAt` helper so both the diffing and exact variants share the same request construction.

**No behavior changes for existing callers** - `UpdateBiosAttributesApplyAt(WithTask)` and `UpdateBiosAttributes` are untouched aside from the internal refactor (verified the ETag `GET` still happens in the same order relative to the empty-payload check as before).

## Test plan
- Added `TestUpdateBiosAttributesApplyAtOmitsUnchangedAttributes` documenting/guarding the existing diffing behavior.
- Added `TestUpdateBiosAttributesExactApplyAt`, `TestUpdateBiosAttributesExactApplyAtWithTask`, `TestUpdateBiosAttributesExact` covering the new methods.
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `go test ./schemas/...` all pass.


schemas/bios.go
```diff
@@ -182,9 +182,16 @@ func (bi *Bios) UpdateBiosAttributesApplyAt(attrs SettingsAttributes, applyTime
 
 // UpdateBiosAttributesApplyAtWithTask is used to update attribute values and set apply time
 // together. It returns a TaskMonitorInfo when the service processes the update asynchronously.
+//
+// Only attributes whose requested value differs from the resource's currently reported value are
```
stmcginnis: Let's try to minimize the Claude-slop here. ;)

stmcginnis:
Thanks for working on this!

Could you actually try your code with the changes in #571? This kind of works around an issue that a caller shouldn't need to worry about. I'm hoping that PR actually addresses things at the source so we don't need to add more surface area and subtle gotchas for users.

mcanevet:
Tried it against real hardware, reproducing the conflict scenario end-to-end (one BIOS write staged, a second hitting the conflict, then recovering by resubmitting the merged attributes). It worked - the merged write landed correctly and applied on reboot. Looks like #571 covers my use case.

stmcginnis:
Excellent - thanks a lot for validating that. I'm going to go ahead with merging that one then. Really appreciate you digging in to this.

PR #572, opened 2026-09-17
fix(schemas): detect collection pagination cycles

ilyam8:
Detect repeated next links in a shared iterative page walker used by both collection APIs. Cyclic responses now return an error instead of repeatedly fetching pages until an external failure, while valid collections remain uncapped and retain existing ordering, partial-result and expansion-fallback behavior.

Tests cover self-links, multi-page cycles, valid257-page collections, later-page failures and expansion fallback.

Full race tests, vet/build, focused Go1.21 tests and Redfish consumer integration passed.


stmcginnis (approved):
Thanks! Seems like maybe a bug in the vendor's implementation that should be addressed, but good to protect against it here. Very slight worry about the extra memory overhead, but I think it would need to be a very broken system for that to really become an issue.

